PR #364, opened 2026-05-16
feat(ts-prompt-assist): B-1b foundation hardening — ts-res integration + 16 PR #362 deferrals + Result-chain refactor

ErikFortune:
Implements B-1b per `.ai/tasks/active/ts-prompt-assist/brief-phase-b-1b.md`. Lands on the cluster integration branch `claude/ts-prompt-assist-features` (NOT release — Guardrail #7).

## Summary

- **Part 1 — ts-res integration.** Replaces the standalone in-record candidate selector with the design §15.5 Option C architecture. `PromptLibrary` holds a long-lived `ResourceManagerBuilder` (parameterized by the new required `qualifiers` create-param), lazy-materializes records via `addLooseCandidate` on first-resolve cache miss (key = `(scope, id, Crc32Normalizer.computeHash(candidates))`, synthesized id = `prompt_<...>`), and drives a `ResourceResolver` over the builder. **Per-resolve flow consumes ts-res's natural priority-descending order (regular instance set preferred when non-empty; default-match set as fallback per Copilot fix `f12a12e6`), collects matching candidates until the first non-partial (full base → stop), then reverses for the join so the base appears first and most-specific overrides last** (per the corrected §10.2 `isPartial` semantic, design amendment `e668e027` — see "Design amendment" below). Trace `candidateMatches[i].matchType` surfaces the real `'match' | 'matchAsDefault'` disposition; `conditions` populated from `resolver.resolveConditionSet(...).matches`. The standalone matcher is REMOVED — no dual-path code.
- **Part 2 — 16 Copilot deferrals from PR #362 closed** (catalog below).
- **Part 3 — Result chaining.** `resolve()` refactored from imperative `isFailure()` cascade to a single fluent chain `walkScopeChain → mergeBindings → materializeIfNeeded → resolveCandidates → renderResolved` (Guardrail #10).

New `IPromptLibraryCreateParams` fields per design §4.1: `qualifiers` (required), `qualifierTypes?`, `cacheListener?`, `logger?` (defaults to `Logging.NoOpLogger`), `resourceBindingDepthLimit?` (default 5; wiring-only until B-2).

## Design amendment (commit `e668e027`)

PR-walkthrough surfaced that the earlier §10.2 walk imported ts-res's `isPartial` flag by name but inverted its semantic — design said *"partial = base layer that participates; non-partial = terminal that stops the walk"*, while ts-res's `resolveComposedResourceValue` treats partial as a more-specific override that layers onto the first full (non-partial) base. Same field, opposite directionality. Same-name-opposite-semantic between two libraries in the same family is a long-tail confusion trap; realigning with ts-res:

- `isPartial: true` = more-specific override
- `isPartial: false` (or omitted) = full base body
- Walk consumes ts-res's natural priority-descending order, collects until first non-partial (full → stop), reverses for the join so base appears first and most-specific override appears last (matching the LLM "later instructions override" convention)
- Hand-rolled "regulars-only preference" filter from the previous Copilot-fix pass removed — ts-res's natural ordering does the right thing once the `isPartial` semantic is right-side-up

`design.md` §5.3 YAML example flipped (base is no-longer-partial; tone / region overrides marked partial). §10.2 algorithm rewritten to document the ts-res alignment. `state.md` records the amendment with rationale. Canonical layering test flipped to match new semantic; bogus "regulars over defaults" test replaced with two clarifying tests covering (a) the regular-base-blocks-default-partial case and (b) a regular-base-plus-regular-partial layering case.

## Closes B-1 deferrals

| # | Item | Status |
|---|---|---|
| 2.1 | `SlotName` Mustache-identifier validation | **Fixed.** `Convert.slotName` now requires `/^[A-Za-z_][A-Za-z0-9_]*$/`. |
| 2.2 | `brandedString` minimal hygiene | **Fixed.** All brands enforce max-length 256 + reject leading/trailing whitespace. `Convert.promptId` additionally rejects `::`. |
| 2.3 | `MustacheTemplateCache` key collision | **Closed by 2.2.** PromptId now rejects `::`, so the flat `${promptId}::${bodyHash}` key is collision-free; comment updated to confirm the dependency. |
| 2.4 | `ConverterRegistry.get` cast surface | **Fixed.** Added kind-verified overload `get(id, kind: T['kind'])` that fails when the registered kind doesn't match the requested kind. No-kind overload kept for the loose case. |
| 2.5 | `promptFileConverter` double-traversal | **Fixed.** Splits `{ candidates, ...descriptorRaw }` upfront so `descriptorConverter` never sees the candidates field. |
| 2.6 | `_slotBindingHolder` mutable-singleton | **Refactored.** Replaced with `_bindingDispatch` carrying an explanatory comment (the closure-thunk alternative trips `@typescript-eslint/no-use-before-define` under our config; the holder pattern is the residual-cleanest path given the mutual recursion). `slotBindingConverter` now uses `Converters.discriminatedObject` per code-reviewer P2-1. |
| 2.7 | `joinBodies` trim regex contract | **Fixed.** Widened to `/\s+$/` so trailing-space-only bodies are also trimmed. |
| 2.8 | `selectCandidates` tiebreak | **Closed by Part 1.** ts-res priority-based tiebreaking now applies. |
| 2.9 | `IQualifierContext` flat-string asymmetry | **Closed as design rationale.** The TSDoc note now explains that ts-res's `SimpleContextQualifierProvider` itself takes a flat record; descriptor-side complexity flows through ts-res's resolver. |
| 2.10 | `PromptStoreFixture` eager `mapResults` | **Fixed.** Three encode steps collapsed into a chained pipeline via dedicated helpers (`encodeRecords`, `encodeBindings`, `encodeQualifiers`) under a single `onSuccess` chain. |
| 2.11 | `chainWalker` sequential `getBindings` | **Fixed.** Now parallelizes via `Promise.all` + `mapResults`. |
| 2.12 | `describe` cache validation + invalidation | **Fixed.** Load-time check rejects descriptors that diverge structurally across scopes (via `Hash.Crc32Normalizer.computeHash`); new `invalidateDescriptor(id)` public method for write-capable adapters. |
| 2.13 | Reserved Windows device set | **Fixed.** Widened to `COM0..9` / `LPT0..9` (the 0-suffixed variants from Windows 11 / Server 2022). |
| 2.14 | `FileTreePromptStore.create` async-sync | **Closed as documentation.** Async signature is forward-compat for real-I/O FsTree adapters; TSDoc note retained without the deferral framing. |
| 2.15 | `bindingMerger` cross-scope intent | **Closed as design rationale per §10.4.** TSDoc note retained. |
| 2.16 | `resolveAndValidateOutput` full-resolve cost | **Closed.** Resolve-first is the correct shape per B-4; deferral annotation removed. |

## Code-reviewer P2 fixes (commit `e5a43c33`)

- **P2-1:** `slotBindingConverter` and `outputConverter` now use `Converters.discriminatedObject` instead of hand-rolled `kind` dispatch.
- **P2-2:** Added multi-scope identical-descriptor test for `describe`; removed the matching c8-ignore directive (success tail is now fully exercised).
- **P2-3:** Acknowledged — `get(id)` no-kind overload's narrowing is intentional and B-4 will add the runtime re-check at the chain runner per §17.2.4.
- **P2-4:** `_addCandidatesToBuilder` now chains via `.asResult` instead of imperative `isFailure()` guard.

## Copilot review fixes (commit `f12a12e6`)

- **Default-match ordering (correctness):** ts-res's `resolveDecision` returns regular and default-match instance indices as separate priority-ordered groups; defaults are strictly less-preferred fallbacks. The prior code concatenated both groups, which let default partials layer before regular terminals. Now picks only the regular set when non-empty; otherwise falls back to defaults. New test confirms a regular base wins over a default-match candidate.
- **Validate-before-mutate:** `_addCandidatesToBuilder` splits into two phases — validate every candidate's decl shape first (no builder mutation), then commit the validated decls sequentially with short-circuit. A malformed record never leaves partially-added candidates in the long-lived builder.
- **TSDoc link target:** dropped the broken `Resources.` qualification on the class-comment `{@link}`.
- **Dead `_descriptorHashes` state:** removed entirely (was never read).

## Out of scope

- B-2 (resource-binding resolve: recursive resolver, cycle detection, depth-cap enforcement, OQ-2 strict-replace, `IResourceBindingTraceEntry.innerTrace`). `serializeResource` continues to fail-loud.
- B-3 (FsTree smoke test; filename-id consistency in `descriptorConverter`; `_qualifiers.yaml` root-level hardening).
- B-4 (`resolveAndValidateOutput` pipeline; input safeguards beyond `'enforced-override-ignored'`).
- B-5 (README, LIBRARY_CAPABILITIES.md, rush change file).

## Test plan

- [x] `rushx build` passes in `ts-prompt-assist`
- [x] `rushx lint` passes in `ts-prompt-assist`
- [x] `rushx test` passes — **82 tests, 100% coverage** on statements / branches / functions / lines
- [x] `rushx fixlint` run before final commit
- [x] api-extractor `etc/ts-prompt-assist.api.md` regenerated
- [x] `code-reviewer` agent invoked before opening this PR; P2 findings applied (commit `e5a43c33`)
- [x] Copilot review pass applied: P1 correctness fixes in `f12a12e6`; design-correction follow-up in `e668e027`
- [x] All `// Copilot review (PR #362, deferred to B-1b)` annotations removed from source
- [x] `state.md` updated with the B-1b deliverable record (incl. design amendment rationale)
- [x] No `any`, no `console.*`, no `node:fs` direct, no `JSON.stringify`-compare, no `Result<void>`, no unsafe casts in product code

https://claude.ai/code/session_01Smb6scXdkqKGnohkQiJrZo

---
_Generated by [Claude Code](https://claude.ai/code/session_01Smb6scXdkqKGnohkQiJrZo)_


libraries/ts-prompt-assist/src/packlets/resolve/promptLibrary.ts
```diff
@@ -229,13 +305,196 @@ export class PromptLibrary<TResponse extends { kind: string } = { kind: string }
...
+  private _renderResolved(
+    req: IPromptResolveRequest,
+    walked: {
+      readonly record: IStoredPromptRecord;
+      readonly winningScope: ScopeKey;
+      readonly scopesConsulted: ReadonlyArray<ScopeKey>;
+    },
+    mergeResult: {
+      readonly merged: ReadonlyMap<SlotName, IBindingTraceEntry>;
+      readonly safeguardFindings: ReadonlyArray<import('../types').ISafeguardFinding>;
```
ErikFortune, line 567: can we skip the dynamic imports, please.

libraries/ts-prompt-assist/src/packlets/types/ids.ts
```diff
@@ -70,20 +70,50 @@ export type AxisName = Brand<string, 'AxisName'>;
...
+  /** When set, additionally rejects the supplied substring inside the
+   *  candidate (e.g. `::` for `PromptId` to defuse the cache-key
+   *  delimiter collision in `MustacheTemplateCache`). */
+  readonly rejectSubstring?: string;
+  /** When set, the candidate must match this regular expression. Used
+   *  for `SlotName` to enforce the Mustache "name" production. */
+  readonly mustMatch?: RegExp;
+}
+
+function brandedString<T extends string>(brand: string, options?: IBrandValidationOptions): Converter<T> {
```
ErikFortune: nit: we might want a less generic name for this since we have lot of branded strings in the system that use totally different semantics - these are branded mustache placeholders, right?

PR #376, opened 2026-05-18
feat(ts-res): qualifier ergonomics PR B — F1 mixed-shape collector + F14 Partial-widened context

ErikFortune:
## Summary

Addresses round-1 pressure-test findings **F1** and **F14** from
`.ai/tasks/active/ts-prompt-assist/pressure-test-findings-round-1.md`.
PR B of the three ergonomics PRs absorbing the ts-prompt-assist pressure-test pressure.

### F1 — `Qualifiers.QualifierCollector.create` accepts mixed-shape input

`qualifiers` now accepts `ReadonlyArray<string | IQualifierDecl>`.

- A string element `'foo'` is sugar for a literal qualifier: the library
  synthesizes a `LiteralQualifierType` named `'foo'` and a decl
  `{ name: 'foo', typeName: 'foo', defaultPriority: <descending> }`.
- **Synthesized priorities are distributed across the
  `[MinConditionPriority..MaxConditionPriority]` range adaptively to input length**
  (`step = max(1, floor((Max - Min) / length))`,
  `priority = max(Min, Max - index * step)`). Earlier-in-array gets higher
  priority. The formula never overflows the `ConditionPriority` range
  regardless of input length; trailing priorities collapse at
  `MinConditionPriority` for length > 1000.
- `qualifierTypes` is now optional, **but only when every element of
  `qualifiers` is a string** (the simpler fallback rule from the brief).
  If any element is a declaration and `qualifierTypes` is omitted, the
  error names the offending typeNames specifically (improvement over
  the previous generic "qualifierTypes must be supplied" message).

The constructor is reshaped from `(params)` to `(qualifierTypes, qualifiers?)`
— protected, no subclasses in the monorepo, no breakage. All decl
normalization runs in `create()` before the constructor is invoked.

### F14 — Widen `qualifierValues` to `Partial<...>`

`ISimpleContextQualifierProviderCreateParams.qualifierValues` and
`IValidatingSimpleContextQualifierProviderCreateParams.qualifierValues`
widen from `Record<...>` to `Readonly<Partial<Record<...>>>`.
This lets callers write `tone === 'formal' ? { tone: 'formal' } : {}`
without an explicit annotation. The loader skips `undefined` values
(semantically equivalent to omitting the key, which is how the candidate
selector treats a missing key).

**Note about F14's surface:** the finding names `IQualifierContext`
"from `@fgv/ts-res`", but that type is actually defined in
`@fgv/ts-prompt-assist` (`packlets/types/qualifiers.ts`). The
equivalent ts-res surface — and the one a real consumer flows through
into ts-res's `SimpleContextQualifierProvider` — is the
`qualifierValues` parameter on the two factories widened here. **PR C
will widen `@fgv/ts-prompt-assist`'s `IQualifierContext` type itself**;
this PR makes the ts-res side of the cascade ready.

### Review-feedback cleanup commit (`1ef0d277e`)

Two changes from orchestrator review:

1. **`_resolveCreateParams` refactored to use `Converters.oneOf`** instead
   of hand-rolled `typeof q === 'string'` filter-and-type-guard. The
   canonical fgv pattern for `string | IQualifierDecl` is
   `Converters.oneOf([Converters.string, qualifierDecl])` + `arrayOf`,
   which validates AND discriminates at the public-API boundary. The
   downstream partition is a single forward pass with natural TS
   narrowing (no unsafe cast). Side win: malformed `IQualifierDecl`
   shapes fail at entry with a clear oneOf error rather than deep inside
   `_qualifierFactory.orThrow()`.

2. **Length-adaptive priority formula** replaces the prior fixed-step
   `(length - index) * 100`. The prior formula overflowed
   `MaxConditionPriority = 1000` at length > 10 (first element would
   have priority 1100, failing validation deep inside the decl converter
   chain). The new formula is described above.

   The previously-introduced `QUALIFIER_COLLECTOR_DEFAULT_PRIORITY_STEP = 100`
   public constant is **removed** — the length-adaptive formula doesn't
   have a meaningful fixed step. The constant was only added in this
   same unmerged PR; removal is pre-publish and non-breaking.

### What's not in scope

- F1's richer constrained-values shape (`{ name, values: ['formal', 'casual'] }`)
  — explicitly deferred to v0.2 per Erik's commissioning note.
- `PromptLibrary.create.qualifiers` mixed-shape cascade on ts-prompt-assist — PR C.
- A standalone `Qualifiers.simpleLiterals(names)` helper — folded into
  the mixed-shape `create` extension.
- Any qualifier-runtime behavior change. This is a constructor /
  decl-shape change only.

### Pre-PR gates (all green)

- [x] `rush build --to @fgv/ts-prompt-assist` — green (ts-res + ts-prompt-assist build clean against the new surface)
- [x] `rushx lint` in `libraries/ts-res` — clean
- [x] `rushx fixlint` was run before the final commit (import block reformatted)
- [x] `rushx test` in `libraries/ts-res` — green, **100% coverage on every file**
- [x] `rushx test` in `libraries/ts-prompt-assist` — pre-existing consumer tests still green
- [x] api-extractor regenerated; `etc/ts-res.api.md` diff included in the commit
- [x] Rush change file under `common/changes/@fgv/ts-res/` — bump type **minor** (additive surface)
- [x] No `any`; all new fallible operations return `Result`; chains throughout

### api-extractor diff summary

- `IQualifierCollectorCreateParams.qualifierTypes`: `ReadOnlyQualifierTypeCollector` → optional
- `IQualifierCollectorCreateParams.qualifiers`: `IQualifierDecl[]` → `ReadonlyArray<string | IQualifierDecl>`
- `QualifierCollector` constructor signature change (protected, additive externally)
- `ISimpleContextQualifierProviderCreateParams.qualifierValues`: `Record<...>` → `Readonly<Partial<Record<...>>>`
- `IValidatingSimpleContextQualifierProviderCreateParams.qualifierValues`: `Record<...>` → `Readonly<Partial<Record<...>>>`

## Test plan

- [x] `Qualifiers.QualifierCollector.create({ qualifiers: ['language', 'tone'] })` synthesizes literal types with length-adaptive descending priorities (1000, 500 for length=2)
- [x] Mixed `['greeting', { name: 'language', typeName: 'language', defaultPriority: 500 }]` with `qualifierTypes` supplied works (greeting → priority 1000, language → explicit 500)
- [x] String element whose name matches an existing type in `qualifierTypes` does NOT duplicate-add
- [x] Decl present + `qualifierTypes` omitted fails with offending typeNames in the message
- [x] Empty inputs (no qualifiers / empty array) succeed with an empty collector
- [x] Existing decl-array consumers (ts-prompt-assist tests, ts-res tests) unaffected
- [x] `qualifierValues: { tone: undefined }` is accepted and the key is skipped on load
- [x] **Length=11 regression test** — verifies the length-adaptive formula stays in range (first priority 1000, last 100); under the prior fixed-step formula the first would have been 1100 and failed validation
- [x] **Malformed `IQualifierDecl` rejected at the Converter boundary** — exercises the new oneOf entry validation

## Reader-side breakage from F14

No reader-side breakage was observed in `@fgv/ts-res`'s own tests or in
`@fgv/ts-prompt-assist`'s tests. The `Object.entries(qualifierValues)`
loop in `SimpleContextQualifierProvider`'s constructor was the only
in-repo reader; it now skips `undefined` values.

https://claude.ai/code/session_01VDi6fazcgVBqaPhbP8nphe

---
_Generated by [Claude Code](https://claude.ai/code/session_01VDi6fazcgVBqaPhbP8nphe)_

### DIFF
--- a/libraries/ts-res/src/packlets/qualifiers/qualifierCollector.ts
+++ b/libraries/ts-res/src/packlets/qualifiers/qualifierCollector.ts
@@ -24,6 +24,8 @@ import * as Common from '../common';
 import {
   captureResult,
   Collections,
+  Converter,
+  Converters,
   failWithDetail,
   mapResults,
   Result,
@@ -33,7 +35,7 @@ import {
   ValidatingConvertingCollector
 } from '@fgv/ts-utils';
 import { IQualifierDecl, IValidatedQualifierDecl } from './qualifierDecl';
-import { QualifierName } from '../common';
+import { MaxConditionPriority, MinConditionPriority, QualifierName } from '../common';
 import { Qualifier } from './qualifier';
 import { IQualifierDeclConvertContext, qualifierDecl, validatedQualifierDecl } from './convert';
 import {
@@ -41,6 +43,18 @@ import {
   QualifierTypeCollector,
   ReadOnlyQualifierTypeCollector
 } from '../qualifier-types';
+
+/**
+ * Converter for a single entry in the mixed-shape
+ * {@link Qualifiers.IQualifierCollectorCreateParams.qualifiers | qualifiers}
+ * input — either a bare axis-name string or a full
+ * {@link Qualifiers.IQualifierDecl | declaration}. Validates the entry
+ * shape at the public-API boundary rather than deferring to the
+ * internal qualifier factory.
+ */
+const _qualifierEntriesConverter: Converter<ReadonlyArray<string | IQualifierDecl>> = Converters.arrayOf(
+  Converters.oneOf<string | IQualifierDecl>([Converters.string, qualifierDecl])
+);
 
 /**
  * Readonly version of {@link Qualifiers.QualifierCollector | QualifierCollector}.
@@ -68,13 +82,6 @@ export interface IReadOnlyQualifierCollector extends Collections.IReadOnlyValida
 }
 
 /**
- * Default priority step used when synthesizing {@link Qualifiers.IQualifierDecl | declarations}
- * from a bare axis-name string in the mixed-shape {@link Qualifiers.QualifierCollector.create | create} input.
- * @public
- */
-export const QUALIFIER_COLLECTOR_DEFAULT_PRIORITY_STEP: number = 100;
-
-/**
  * Parameters for creating a new {@link Qualifiers.QualifierCollector}.
  * @public
  *
@@ -84,9 +91,16 @@ export const QUALIFIER_COLLECTOR_DEFAULT_PRIORITY_STEP: number = 100;
  * sugar for "literal qualifier with this name"; the library synthesizes a
  * {@link QualifierTypes.LiteralQualifierType | LiteralQualifierType} and a
  * declaration `{ name, typeName: name, defaultPriority: <descending> }`.
- * Priorities for synthesized declarations follow `(arr.length - index) * step`
- * (step = {@link Qualifiers.QUALIFIER_COLLECTOR_DEFAULT_PRIORITY_STEP | QUALIFIER_COLLECTOR_DEFAULT_PRIORITY_STEP}), so earlier
- * elements get higher priority.
+ *
+ * Synthesized priorities are distributed across the
+ * `[MinConditionPriority..MaxConditionPriority]` range adaptively to the
+ * array length: `step = max(1, floor((Max - Min) / length))`,
+ * `priority = max(Min, Max - index * step)`. Earlier elements get higher
+ * priority. The formula never overflows the `ConditionPriority` range
+ * regardless of input length; for inputs whose length exceeds the range
+ * cardinality (\>1000 axes), trailing priorities collapse at
+ * `MinConditionPriority` — supply explicit priorities in that regime
+ * if uniqueness matters.
  *
  * `qualifierTypes` may be omitted only when every element of `qualifiers`
  * is a string. If any element is an {@link Qualifiers.IQualifierDecl},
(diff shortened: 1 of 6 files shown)


libraries/ts-res/src/packlets/qualifiers/qualifierCollector.ts
```diff
@@ -97,29 +133,169 @@ export class QualifierCollector
...
+        return succeed({ qualifierTypes: QualifierCollector._emptyQualifierTypes(), qualifiers: undefined });
+      }
+      return succeed({ qualifierTypes: params.qualifierTypes, qualifiers: undefined });
+    }
+
+    const stringNames = qualifiers.filter((q): q is string => typeof q === 'string');
+    const allStrings = stringNames.length === qualifiers.length;
+
+    if (!allStrings && params.qualifierTypes === undefined) {
+      const declTypeNames = qualifiers
```
ErikFortune: We should be using the define converters for the qualifier type decl and for qualifier names here instead of rolling our own.

libraries/ts-res/src/packlets/qualifiers/qualifierCollector.ts
```diff
@@ -97,29 +133,169 @@ export class QualifierCollector
...
+  private static _normalizeQualifierDecls(
+    qualifiers: ReadonlyArray<string | IQualifierDecl>
+  ): ReadonlyArray<IQualifierDecl> {
+    const step = QUALIFIER_COLLECTOR_DEFAULT_PRIORITY_STEP;
+    return qualifiers.map((q, index) => {
+      if (typeof q === 'string') {
+        return {
+          name: q,
+          typeName: q,
+          defaultPriority: (qualifiers.length - index) * step
```
ErikFortune: should probably use a dynamic step here if the default step would end up with priorities < 0

PR #385, opened 2026-05-18
feat(ts-prompt-assist): F1 + F2 + F6 round-2 ergonomics absorb

ErikFortune:
## Summary

Absorbs three findings from the ts-prompt-assist round-2 pressure test (`.ai/tasks/active/ts-prompt-assist/pressure-test-findings-round-2.md`) into the library surface, before sample-app PR #384 rebases + lands.

### F1 — Typed candidate `conditions` keys
- `IPromptStoreFixtureSeed<TQualifierNames>` + `IPromptStoreFixtureSeedRecord<TQualifierNames>` parameterized on `TQualifierNames extends string = string`.
- New `ITypedConditionSetDecl<T>` narrows the KEY side of the seed-side `conditions` field to `T` while preserving ts-res's value-side expressivity (string sugar, `IChildConditionDecl` record-with-details, and the array form). Candidates use the new sibling `ITypedPromptCandidateRecord<T>` whose `conditions` is `ITypedConditionSetDecl<T>`.
- Runtime / store `IPromptCandidateRecord` is unchanged (still string-keyed) — typing is purely compile-time authoring discipline at the seed.
- `PromptStoreFixture.build<TQualifierNames>(seed)` infers the parameter from a typed seed.
- Default `TQualifierNames = string` keeps every existing seed compiling unchanged.
- Verified by a `@ts-expect-error` test on the typo branch: `{ tonr: 'formal' }` against `IPromptStoreFixtureSeed<'tone'>` is now a compile error instead of silently falling through to the base candidate at resolve time.

### F2 — `buildSimpleDescriptor` helper
- New exported function `buildSimpleDescriptor({ id, title, surface?, description? })` returns a fully-shaped `IPromptDescriptor` for the trivial chat case (`schemaVersion: '1'`, `surface: 'chat'`, `slots: []`, `output: { kind: 'free-text' }`).
- Deliberately limited to free-text output — `descriptor.output.kind` is load-bearing (`resolveJsonOutput` / `resolveFreeTextOutput` runtime-verify it), so silently defaulting it for omitted-output prompts would route JSON consumers into the wrong dispatcher. JSON-output prompts continue to author the full `IPromptDescriptor` shape so the `output.converterId` discriminator stays explicit at the call site.

### F6 — README React-wiring section
- `ChatInner` snippet extended with `ChatTone = 'neutral' | 'formal'`, a `<select>` bound to it, and the natural `tone === 'neutral' ? {} : { tone }` qualifier-context wire-through.
[...]

PR #386, opened 2026-05-18
feat(ts-res): parameterize ConditionSetDecl family on TQualifierNames

ErikFortune:
## Summary

Parameterizes ts-res's `ConditionSetDecl` family on a `TQualifierNames extends string` literal-string union. Narrows the axis-name (`qualifierName` on the array form; record keys on the record form) so typo'd axis names fail at compile time when consumers author conditions in code.

**In-place parameterization** (vs. introducing parallel `Typed*` sibling types). Per Erik's adjudication: with default-string the change is shape-equivalent for the array form; the record form's `Partial` widening is the F14-class strict-improvement type-tightening, not a true break.

## Rationale

Round-2 pressure-test finding F1 in `@fgv/ts-prompt-assist` surfaced that candidate `conditions` keys aren't typed against the library's qualifier-name union — same silent-failure mode F3 absorbed for the resolve side, still present on the authoring side.

Per `/published-primitives-reflex`: the typed-condition pattern is generic to anyone authoring ts-res candidates in code. Belongs at ts-res, not at ts-prompt-assist. Mirrors the F1 mixed-shape qualifiers cascade from PR #376 — surface the friction at the consumer layer, fix the primitive at the source.

## What changed

```ts
export interface ILooseConditionDecl<TQualifierNames extends string = string> {
  qualifierName: TQualifierNames;
  // ...other fields unchanged
}

export type ConditionSetDeclAsArray<TQualifierNames extends string = string> =
  ReadonlyArray<ILooseConditionDecl<TQualifierNames>>;

export type ConditionSetDeclAsRecord<TQualifierNames extends string = string> =
  Readonly<Partial<Record<TQualifierNames, string | IChildConditionDecl>>>;

export type ConditionSetDecl<TQualifierNames extends string = string> =
  | ConditionSetDeclAsArray<TQualifierNames>
  | ConditionSetDeclAsRecord<TQualifierNames>;
```

Value-side expressivity fully preserved — string sugar, `IChildConditionDecl` (record-with-details), and the array-form's `ILooseConditionDecl` all continue to be accepted.

## Compatibility

**Array form**: shape-equivalent with `TQualifierNames = string`.

**Record form**: switches from `Record<string, V>` to `Readonly<Partial<Record<TQualifierNames, V>>>`. The `Partial` widening adds `| undefined` to the value type at the index-signature level when `TQualifierNames = string`. This is the F14-class strict type-system tightening — runtime semantics of `Record<string, V>` already include missing-key-produces-undefined behavior; the `Partial` makes TS aware of what's already true.

**One reader-side adjustment was required**: `ConditionSet.getKeyFromLooseDecl` at `conditionSet.ts:200-225` previously assumed `Object.entries` produced non-undefined values; now skips entries where the value is explicitly `undefined` (semantically equivalent to omitting the key). Pre-existing tests pass unchanged with 100% coverage held.

**No `Typed*` parallel types introduced** — ts-prompt-assist's PR #385 will port to drop its local `ITypedConditionSetDecl` and reference `ResourceJson.Json.ConditionSetDecl<T>` directly.

## Pre-PR gate

- [x] `rush build --to @fgv/ts-prompt-assist` green (ts-res + ts-prompt-assist both build clean against the new surface)
- [x] `rushx lint` clean in both packages
- [x] `rushx test` in both packages — 100% coverage held on every file
- [x] api-extractor regenerated (`etc/ts-res.api.md` shows parameterized signatures; no new public exports)
- [x] Rush change file under `common/changes/@fgv/ts-res/` — bump type `minor`
- [x] PR targets `claude/ts-prompt-assist-features` (integration branch)

## Next steps

After this merges: port PR #385 to drop its local `ITypedConditionSetDecl`; reference `ResourceJson.Json.ConditionSetDecl<T>` directly. `ITypedPromptCandidateRecord<T>` stays in ts-prompt-assist (it's a ts-prompt-assist-domain concept). PR #384 rebases onto post-port HEAD; cluster close follows.

https://claude.ai/code/session_01VDi6fazcgVBqaPhbP8nphe

### DIFF
--- a/common/changes/@fgv/ts-res/chore-ts-res-typed-condition-set-decl_2026-05-18-19-14.json
+++ b/common/changes/@fgv/ts-res/chore-ts-res-typed-condition-set-decl_2026-05-18-19-14.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Add typed counterparts to ConditionSetDecl family — TypedConditionSetDeclAsArray, TypedConditionSetDeclAsRecord, TypedConditionSetDecl — parameterized on TQualifierNames for compile-time axis-name discipline in in-code condition authoring; ts-prompt-assist will consume.",
+      "type": "none",
+      "packageName": "@fgv/ts-res"
+    }
+  ],
+  "packageName": "@fgv/ts-res",
+  "email": "noreply@anthropic.com"
+}
--- a/libraries/ts-res/etc/ts-res.api.md
+++ b/libraries/ts-res/etc/ts-res.api.md
@@ -588,7 +588,7 @@ class ConditionCollector extends ValidatingCollector<Condition> {
 // Warning: (ae-unresolved-link) The @link reference could not be resolved: This type of declaration is not supported yet by the resolver
 //
 // @public
-const conditionDecl: ObjectConverter<ILooseConditionDecl, unknown>;
+const conditionDecl: ObjectConverter<IConditionDecl, unknown>;
 
 // @public
 export type ConditionIndex = Brand<number, 'ConditionIndex'>;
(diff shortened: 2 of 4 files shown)


ErikFortune:
Superseded by the B+C ts-res stream (commission in flight).

The decision-track is captured in the design-options brief and evaluation doc now landed on integration (PRs #387, #388, #389). Headline finding: this PR is leaf-only — the cascade doesn't reach the container types (`IChildResourceCandidateDecl`, `IResourceCollectionDecl`, `IResourceTreeRootDecl`), so `TQualifierNames` has no flow path from a consumer's authoring chain into a `ConditionSetDecl` shape. Combined with the cast-temptation critique (typed authoring without runtime teeth pressures consumers toward `as` casts on round-trip flows), the right shape is B+C together: cascade through containers AND parameterize the Converter pipeline against a `QualifierCollector`-derived literal set.

The `Partial` widening fix (the genuine latent-bug improvement here — `ConditionSet.getKeyFromLooseDecl` handling explicit-`undefined` record entries) gets carried into the B+C stream's scope.

Closing.

---

PR #497, opened 2026-06-25
feat(ts-agent-memory): B1 store + index packlets

ErikFortune:
## Summary

Implements the **B1 tier** of `@fgv/ts-agent-memory`: the `store` and `index` packlets — a FileTree-backed, writable, content-hash-deduped memory store plus the derived in-memory index it maintains. This is the first tier with real I/O; **all I/O goes through FileTree, never `fs`.** Built on the merged B0 (`types` + `converters`); no B0 file was modified.

Targets the `ts-agent-memory` integration branch (not `release`).

## What's in scope

**`store/`**
- `FileTreeMemoryStore` (`IMemoryStore`) over `FileTree.IMutableFileTreeDirectoryItem`. `put` sequence: validate body via the registry → compute a `Crc32Normalizer` content hash over `{ kind, body, links }` → **scope-wide dedup BEFORE policy** → `IWritePolicy.admit` (accept / reject / cull-oldest) → `applyUpdate` on update → store-stamped `created` / `updated` / `seq` (global counter incremented inside the write-lock) → `serializeMemoryFile` → FileTree write → `index.patch`. `get(kind, entityId)` routes through the kind's `IIdentityCodec`; plus `getById(scope, id)`, `list(filter)`, `delete(kind, entityId)`.
- `verifyFilenameId` (id ↔ filename + codec round-trip) on every load.
- **Flat (non-versioned) layout only** — a codec reporting `isVersioned: true` fails loudly (versioned/temporal write path is a fast-follow). Non-string bodies also fail loudly (the on-disk format is YAML frontmatter + markdown body).
- Per-instance async **write-lock** serializing `put`/`delete`; initial FileTree walk on `create()` rebuilds the index and resumes `seq` past the highest persisted value.
- `defaultMemoryScopeEncoding`: multi-segment (`/`-separated), each component validated portable-filename-safe (so the Phase-C MTM `conversations/<id>` scope is additive).

**`index/`**
- `MemoryIndex` (`IMemoryIndex`): derived `byKind` / `byTag` / `byRecency` / `backlinks` maps; `rebuild(entries)` + incremental `patch(op, entry)`. A pure, rebuildable in-memory derived view — never the source of truth.
- **Acyclic store↔index wiring:** the store depends on the index interface; the index imports only `types/` and never the store (the store does the FileTree walk and hands the index a record list).

## Out of scope (later tiers)
`retrieve` / `observe` / `vector` packlets (B2); LTM/MTM identity codecs + `MemoryCapCullPolicy` (Phase C); versioned/temporal write path; `watch`; L2 tools; L3 ingest.

## Testing & gates
- `rushx build` ✅, `rushx lint` ✅ (0 warnings), `rushx test` ✅ **100% coverage** (stmts/branch/funcs/lines), `rushx fixlint` run pre-commit, api-extractor report regenerated + committed.
- Tests use an **in-memory FileTree only** (no real `fs`): dedup no-op, LWW update via `applyUpdate`, codec-routed `get`, `verifyFilenameId` mismatch, multi-segment scope encoding, `isVersioned` loud-fail, cull-oldest, write-lock serialization, index rebuild + incremental patch.
- No `any`; all fallible ops return `Result` (no `Result<void>`); all I/O via FileTree.

## Review notes (layer 1 — internal `code-reviewer` pass)
Ran the `code-reviewer` agent on the final diff before opening this PR. **No P1.** P2/P3 resolved or dispositioned:
- **Fixed:** `MemoryIndex._keyOf` composite-key separator made an explicit `\0` escape (collision-proof, matches doc); `_writeFile` return type `Result<string>` → `Result<true>`; dedup comment corrected to the real hash domain (`{kind, body, links}`); removed a dead `segments.length === 0` branch in `defaultMemoryScopeEncoding`.
- **Dispositioned (deferred):** internal sync `Result` chaining off the `async put` is sound while the FileTree read API is synchronous; migrating `_readRecord`/`_writeResolved`/`_evict` to `AsyncResult` when a genuinely-async adapter is wired is a fast-follow (the public `IMemoryStore` surface is already `Promise<Result<…>>`).

## Design-fidelity note for reviewers
`contentHash` covers `{ kind, body, links }` per design-lock §2.5 (excludes `tags`/`provenance`). Combined with scope-wide dedup this means (a) a **tags-only** re-put collapses to a no-op and (b) two different entities with identical `{kind, body, links}` in a scope collapse — both tested and faithful to the locked spec; flagged here in case Phase C wants the hash domain widened (would be a design-lock amendment).

---
_Generated by [Claude Code](https://claude.ai/code/session_01FWXjEN1PJTPs6MuC65E2zh)_

## Summary by CodeRabbit

* **New Features**
  * Added memory indexing for faster lookups, including recency sorting, tag filtering, and backlink tracking.
  * Added a file-backed memory store that persists records to disk and keeps the index in sync.
  * Added scope/path encoding with validation for safer nested storage locations.
* **Bug Fixes**
  * Improved update behavior to preserve creation time while consistently advancing updated timestamps and sequence numbers.
  * Strengthened validation and loud failure modes for mismatched/unsupported data, with correct dedup and delete behavior.
* **Tests**
  * Expanded unit coverage for indexing, persistence, encoding, dedup, policies, concurrency, and error cases.


ErikFortune:
@coderabbitai review

---

ErikFortune:
Thanks for the review — round-1 findings addressed in `3e8d21cc`:

- **Backlinks scope-awareness** (✅ fixed): `_backlinks` now keys source membership by the composite `(scope, id)` instead of the bare `MemoryId`, so two same-id sources across scopes are tracked independently and removing one no longer drops the other's inbound edge. Added the requested same-id-across-scopes backlinks test.
- **Persist the policy-updated body** (✅ fixed): the update path now persists `applyUpdate`'s returned `body` and **recomputes `contentHash` over the policy's actual `{ kind, body, links }`**, with a loud guard if a policy returns a non-string body. (No behavior change for LWW; added a non-string-body test.)
- **Cull-oldest deletes before persist** (✅ fixed): reordered so the new record is built/serialized/written **before** any eviction runs — existing records are never deleted until the replacement is durably on disk.
- **Result matchers in test assertions** (✅ fixed): converted the flagged assertion-context `.orThrow()` extractions to `toSucceedAndSatisfy(...)` / `toSucceedWith(...)`.

**Dispositioned — do not dedup on CRC32 alone:** declining this one. CRC32 is the **design-locked** content-hash dedup key for this package (design-lock §2.5/§4.2), so hash-equality-as-dedup is the specified behavior rather than an oversight. Adding a canonical-tuple confirmation on a hash hit would introduce a `hash-matches-but-content-differs` branch that can't be covered without synthesizing a CRC32 collision, which would break the package's 100%-coverage gate. Widening to a stronger/longer hash is a design-lock amendment and is out of scope for this B1 tier — noting it for follow-up rather than changing it here.

Gates green: build, lint, **100% coverage (152 tests)**, no public-API change.

@coderabbitai review

---

PR #504, opened 2026-06-29
fix(ts-extras/ai-assist): correlate client-tool continuation ids to the buffered tool_use id

ErikFortune:
## Summary

Fixes the production bug reported by PersonAIlity: Anthropic intermittently returns
`invalid Prompt - malformed identifier`, **only** on turns where the model calls a client tool
(more tools offered → more frequent). A direct-answer turn on the same agent/model succeeds.

This is a **bug-fix + hardening + root-cause** change to `@fgv/ts-extras`'s `ai-assist` packlet
(active surface). It is **additive / type-only at the public boundary** — the api-extractor report
is unchanged (`executeClientToolTurn`'s signature is untouched; the affected builders are
internal-only exports). No behavior change beyond the id-correlation fix + the diagnostic.

## Root cause

The per-provider continuation builders never enforced that each `tool_result.tool_use_id`
(Anthropic) / `function_call_output.call_id` (OpenAI) is drawn from the set of assistant
`tool_use.id`s actually present in the accumulation buffer. They keyed the id off the tool-result
record with `tool_use_id: r.callId ?? r.toolName`. Two compounding defects:

1. **Name fallback (`?? r.toolName`)** — when `callId` is nullish, the builder emitted the tool
   *name* as the id. A tool name is never a `toolu_*` / `call_*` id, so the provider rejects it as
   a "malformed identifier".
2. **`??` does not treat empty-string as missing** — an empty id flows straight through as
   `tool_use_id: ''`, also malformed, with no fallback even firing.

Because the builder trusted `callId` blindly and never cross-checked it against the buffered
assistant `tool_use.id`s, **any** upstream loss/divergence of the id surfaced as a malformed wire
body at the provider instead of a loud, field-attributable failure in our code. The identified
upstream divergence path is the **silent orphaned-block drop** in `anthropic.ts`: a `tool_use`
`content_block_start` that failed the `block.id && block.name` guard was dropped from the buffer
with no diagnostic, and its subsequent `input_json_delta` chunks were then silently ignored.

**Honest scope note on the live trigger:** within a single well-formed Anthropic stream the
adapter's own guard means `callId` always equals a buffered id, so the pure single-turn path
can't fire the bug. The decisive end-to-end-reproducible divergence is **buffer-index reuse**
(captured as a test). Rather than pin a single live trigger speculatively, the fix makes every
such loss loud and impossible to mis-key, and adds a field-confirmable diagnostic so the consumer
can corroborate on a live failing turn by capturing the request body. Full write-up in
`.ai/tasks/active/ai-assist-client-tool-id-fix/state.md`.

## Fix

1. **Single source of truth.** `buildAnthropicContinuation` / `buildOpenAiContinuation` now return
   `Result<IAiClientToolContinuation>` and validate that every result's `callId` is non-empty
   **and** present in the set of buffered assistant tool_use ids; otherwise they fail loud (naming
   the tool + offending id). `tool_use_id` / `call_id` is taken from the verified `callId` —
   **never** `toolName`.
2. **Empty-string = missing.** New `isUsableId(id)` predicate replaces every `??`-based id check.
3. **No silent drop.** `anthropic.ts` now emits a loud `logger.warn` (new stable
   `MALFORMED_TOOL_USE_WARN_TAG = 'ai-assist:malformed-tool-use'`) for an orphaned `tool_use`
   block instead of dropping it.
4. **OpenAI parity** — same correlation on the Responses path.
5. **Decisive diagnostic** — both builders log the outgoing id↔id pairing at `detail` (debug) level
   when a logger is supplied; `executeClientToolTurn` threads its logger through and now also yields
   an inline `error` event on a bad correlation (parity with the stream-open-failure path).

The Gemini builder is intentionally unchanged (correlates by tool name by design — Gemini has no
call ids).

## Tests

- **Divergence repro → now fails loud (not mis-keys):** absent / empty-string / mismatched callId,
  and empty buffered tool_use id — each fails loud (replaces the old "uses toolName as fallback"
  tests). OpenAI parity for the same paths.
- **Happy-path id-equality:** `tool_use.id === tool_use_id` for single + parallel calls (parallel
  supplied out of buffer order to prove correlation is by id, not position).
- **Orphaned-block:** missing id / missing name → `logger.warn` with the tag, no tool call, empty
  buffer; plus a no-logger variant.
- **End-to-end fail-loud through `executeClientToolTurn`:** buffer-index-reuse SSE → `nextTurn`
  fails loud and an inline `error` event is emitted.
- **Diagnostic logging** asserted for both builders.

## Gates

- `rushx build` ✅ · `rushx lint` ✅ (after `fixlint`) · `rushx test` ✅ — **100%** statements /
  branches / functions / lines in `@fgv/ts-extras`.
- api-extractor report unchanged → no public-surface change.

## Review loop

- **Internal `code-reviewer` pass run on the final diff before opening this PR.** Findings: **no
  P1**, two P2, two P3.
  - P2 (no `error` event yielded on continuation failure) — **fixed** (now yields inline error).
  - P2 (add `c8 ignore` to the empty-id buffer guard) — **dispositioned: declined**; the branch is
    *tested* and coverage is 100% without a directive (directives are for unreachable code).
  - P3 (unsafe `as string` in a test helper) — **fixed** (typeof guard).
  - P3 (diagnostic could log an empty pairing for a direct caller) — **dispositioned: declined**;
    unreachable via `executeClientToolTurn`, guarding would add an untested branch.

## Scope held

Out: the Gemini builder (name-correlated by design); any alias refactor (separate stream).

---
_Generated by [Claude Code](https://claude.ai/code/session_01P2EBCaYjg9YveM8NPsEFLr)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool-call correlation during AI-assisted streaming so matching IDs are required and mismatches now fail clearly instead of falling back silently.
  * Malformed tool-use blocks now trigger a visible warning and no longer disappear without notice.
  * Continuation errors are now surfaced inline in the event stream, making failures easier to detect and debug.
  * Added stronger validation for parallel and single tool-call flows to prevent incorrect tool-result pairing.

### DIFF
--- a/.ai/tasks/active/ai-assist-client-tool-id-fix/state.md
+++ b/.ai/tasks/active/ai-assist-client-tool-id-fix/state.md
@@ -1,0 +1,118 @@
+# ai-assist-client-tool-id-fix — state
+
+**Library:** `@fgv/ts-extras` (`ai-assist` packlet, active surface)
+**Branch:** `claude/client-tool-id-mismatch-leu8do` (off `release`) — PR target `release`
+**Type:** bug-fix + hardening + root-cause. No public API surface change (api-extractor report unchanged).
+
+## Symptom (consumer: PersonAIlity)
+
+Anthropic intermittently returns `invalid Prompt - malformed identifier`, **only** on turns
+where the model calls a client tool. More tools offered → more frequent. Direct-answer turns
+on the same agent/model succeed.
+
+## Root cause
+
+The per-provider continuation builders never enforced that each
+`tool_result.tool_use_id` (Anthropic) / `function_call_output.call_id` (OpenAI) is drawn from
+the set of assistant `tool_use.id`s actually present in the accumulation buffer. They
+independently keyed the id off the tool-result record with
+`tool_use_id: r.callId ?? r.toolName`. Two compounding defects make that catastrophic rather
+than merely fragile:
+
+1. **Name fallback (`?? r.toolName`).** When `callId` is nullish, the builder emitted the tool
+   *name* as the id. A tool name is never a `toolu_*` / `call_*` id, so the provider rejects it
+   as a "malformed identifier." (`clientToolContinuationBuilder.ts` Anthropic ~line 151, OpenAI
+   ~line 219.)
+2. **`??` does not treat empty-string as missing.** An empty id flows straight through as
+   `tool_use_id: ''` — also a malformed identifier — with no fallback even firing.
+
+Because the builder *trusted* `callId` blindly and never cross-checked it against the buffered
+assistant `tool_use.id`s, **any** upstream loss/divergence of the id surfaced as a malformed
+wire body at the provider instead of a loud, field-attributable failure in our code.
+
+### The identified upstream divergence path (anthropic.ts:~333, the silent orphaned-block drop)
+
+`content_block_start` handling buffered a `tool_use` block only when
+`block.type === 'tool_use' && block.id && block.name`. A `tool_use` block_start that failed that
+guard (missing/empty id or name — provider drift or a truncated stream) fell through **silently**:
+no buffer entry, no `client-tool-call-start`, no `logger.warn`. Its subsequent
+`input_json_delta` chunks (same SSE index) then hit `if (block?.type === 'tool_use')` at
+~line 368 with `block === undefined` and were **silently ignored** too. The block simply vanished.
+
+This is a real silent-corruption path: the model's intended `tool_use` is dropped from the
+assistant turn with zero diagnostic. Note it does **not**, by itself, manufacture a
+*name-as-id* tool_result (a dropped block emits no `client-tool-call-done`, so no tool-result is
+produced for it). The decisive, end-to-end-reproducible divergence that *does* yield a
+tool_result whose id is absent from the final assistant turn is **buffer-index reuse**: if two
+`tool_use` blocks share an SSE `index` (e.g. an absent/duplicated `index`, which the payload type
+already marks optional and defaults to `0`), the later block overwrites the earlier one in the
+buffer after the earlier one already emitted its `client-tool-call-done`. The executed result is
+then keyed to an id no longer present in the buffer → the builder used to mask this with the name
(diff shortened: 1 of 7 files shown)


ErikFortune:
@coderabbitai review

---

ErikFortune:
Thanks for the review. Dispositions on the two nitpicks:

**1. Empty-string malformed `tool_use` adapter case — accepted.** Added in `5e50ca00`. Good catch: the adapter-level orphaned-block tests covered missing-id / missing-name but not an empty-string id, so a future refactor of the `block.id && block.name` guard to `!== undefined` would have passed the suite while reintroducing the bug. (The empty-string-is-missing contract was already covered at the *builder* level via `isUsableId`; this closes the gap at the adapter level too.)

**2. `.orThrow()` → `toSucceedAndSatisfy(...)` in the builder tests — declined, with reasons.** The cited lines are all **pre-existing** tests for `buildAnthropicContinuation` / `buildOpenAiContinuation` / `buildGeminiContinuation`. Before this PR the builders returned a plain value; this PR changed two of them to return `Result<T>`, and the minimal adaptation to keep those structural-assertion tests compiling was `.orThrow()`. Converting all ~18 to wrapped matchers is a test-refactor beyond this bug fix's scope, and `clientToolContinuationBuilder.test.ts` is already at the repo's `max-lines: 2000` cap — the wrapper conversion would push it over and force a file split. The repo's own guidance is to avoid refactoring surrounding code during a bug fix. Note the **new** behavior this PR introduces *does* use matchers throughout (`toFailWith`, `toSucceed`, `toSucceedAndSatisfy`). Happy to do the matcher migration + file split as a separate cleanup if the maintainer wants it.

On the **Docstring Coverage** pre-merge warning (71.43% < 80%): the un-documented functions are test-only helpers (`asString`, `toolUseIds`, `runOrphanedToolUse`, etc.); the repo doesn't require TSDoc on test helpers, so treating that as a non-blocking advisory.

Both findings were `🔵 Trivial`, so I'm treating this as the diminishing-returns point for the review loop.

---

PR #528, opened 2026-07-08
test(ts-agent-memory): antagonist torture tests + two store fixes they surfaced

ErikFortune:
## Summary

The **agent-memory antagonist** stream (phase 1): hole-driven adversarial torture tests for the seven near-miss invariant classes that were each caught *late* across the agent-memory arc (L2 tools, annotations, temporal, L3 ingest). The charter was: for each invariant, construct the input that would fail against a plausible-but-wrong implementation, and **if a torture test surfaces a real bug, STOP and report it** rather than paper over it.

It found **two real, confirmed store bugs** on the (corrected, post-L3) substrate. Both are fixed here with regression tests, per the principal's disposition.

### Bug 1 — content-hash dedup swallowed a same-id metadata-only update (all three write paths)

`KnowledgeLwwPolicy` / `MemoryCapCullPolicy` / `TemporalVersionedPolicy` all declare `tags`/`provenance` in their `mutableFields` — "the metadata a consumer may revise without minting a new entity." But the store's content-hash no-op fired **before** the LWW / merge path on a content-hash match alone, so a same-id re-put whose body+links were unchanged while tags/provenance were revised was collapsed as a duplicate — making the declared mutable surface unreachable.

- **Flat path (`_writeResolved`)**: content collapse now fires only across a **different** id (`_findByContentHash` gained an `excludeId`); a same-id re-put is a no-op only when the mutable metadata (`{tags, provenance}`, canonicalized via the CRC normalizer) is **also** unchanged, else it reaches `applyUpdate`. Applies uniformly to both content- and entity-dedup scopes.
- **Temporal path (`_putVersioned`)** *(caught by the internal code-reviewer as an incomplete first fix)*: the same gate was missing here, and worse — `_buildVersionedRecord` calls `applyUpdate` specifically to mint a new version carrying merged metadata, but the early dedup return made that branch unreachable, so a metadata-only revision on a temporal record was **dropped entirely**. The gate now lets a metadata-only revision mint a new version via the existing merge path.

Cross-id content collapse and identical-re-put idempotence are preserved (regression-tested in all directions). `embeddingRef` is deliberately excluded from the metadata-equality check (it is store-derived, not caller metadata — including it would break the no-op once an embedder is wired).

### Bug 2 — tampered `envelope.entityId` loaded undetected

`_verifyLoaded` checked `envelope.id === filename stem` and the codec round-trip, but the round-trip derives `entityId` from the **scope path** and never read the envelope's own `entityId` field — so a corrupt/tampered file whose frontmatter declared a foreign `entityId` loaded with the wrong id, which is then trusted verbatim downstream (e.g. merge-into re-addressing). `_verifyLoaded` now cross-checks `envelope.entityId` against `codec.decode(scope, stem)` and rejects a mismatch. Verified safe for all four codecs (flat Knowledge/Ltm brand the stem, Temporal decodes the base id, MTM reconstructs the `conv:turn` composite) — no false-rejection of valid records.

## Torture tests (test-only, all genuinely adversarial)

Each documents the wrong-implementation it guards:
- `antagonistTemporalBoundary.test.ts` — inclusive `valid_at` start boundary (`>` vs `>=`); out-of-order backdated writes closing prior intervals at the next version's `valid_at`, not `now`.
- `antagonistCycleAndParity.test.ts` — a proposed edge closing a cycle **through a pre-existing on-disk edge** (guards an orchestrator that forgets the store snapshot); `cycleGuard: 'off'` actually disables; enum-branch parity across all three target-bearing verdicts.
- `antagonistRoundTrip.test.ts` — whole-envelope `toEqual` with every optional field populated (guards silent field-drop) + absent-stays-absent (null/undefined drift).

## Review

- **Internal `code-reviewer` pass** run pre-PR on the full diff. Findings: **P1** — Bug 1's fix was incomplete on the temporal path (fixed, above); **P2** — no entity-scope regression test for the shared gate (added); **P3** — inline dynamic `import()` in a torture test (converted to static). Both stated fixes independently verified sound (same-id/cross-id split, `''` sentinel safe in both failure directions, Bug 2 cross-check safe for all four codecs). All torture tests judged genuinely adversarial, not dressed-up happy paths.

## Gates

`rushx build` ✅ · `rushx lint` ✅ (clean, `fixlint` run) · `rushx test` ✅ **100% coverage** (statements/branches/functions/lines). Additive only — no public API surface change (both fixes touch private methods); `api.md` unchanged. Changeset entry: `patch` (production bug fixes).

---
_Generated by [Claude Code](https://claude.ai/code/session_01T4rVgY8NFE7zNBNLgEVCch)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Metadata-only updates now correctly create a new version or update the existing record instead of being ignored.
  - File loading now rejects records when the stored identifier doesn’t match the file’s scoped identifier.
  - Bi-temporal date handling now respects inclusive start and exclusive end boundaries.
  - Ingest checks now correctly block cycles using existing stored links and handle resolver targets more consistently.

- **Tests**
  - Added broader coverage for serialization, temporal behavior, data integrity, and ingest edge cases.

### DIFF
--- a/.ai/tasks/active/agent-memory-antagonist/brief.md
+++ b/.ai/tasks/active/agent-memory-antagonist/brief.md
@@ -1,0 +1,68 @@
+# Brief — agent-memory antagonist: adversarial torture tests for the near-miss invariant classes
+
+> **STATUS: DRAFT — commission AFTER `agent-memory-l3-ingest` (#527) merges to `release`.**
+> The antagonist tortures the *corrected* substrate, not a known-broken merge-into path.
+> Scope v1: `@fgv/ts-agent-memory` (temporal + L3 ingest + L2 write path). A sibling
+> `ai-assist-antagonist` (provider response-shape + param-rejection class) is a natural follow-on.
+
+**Surface:** `@fgv/ts-agent-memory` (**active** — additive test-only work; no production behavior change unless a torture test surfaces a real bug, in which case STOP and report it, don't paper over it).
+**Ships under the enforced coverage gate.** New tests only; existing 100% coverage must not regress.
+
+## Charter — hole-driven, not coverage-driven
+
+You are an **antagonist**, not an SDET writing happy-path coverage. Your mandate: **assume every invariant below is subtly violable, and for each, construct the minimal input or sequence that breaks it, then write the test that would FAIL against a plausible-but-wrong implementation.** A torture test that passes on the first run is only interesting if you can articulate the wrong implementation it would have caught. Where a test surfaces a *real* current bug: **STOP, do not fix it silently, report it** (per TESTING_GUIDELINES "never paper over failures") — the point of this stream is to find the bugs the happy-path suite missed.
+
+**Output form is open — choose per target:**
+- **Adversarial unit tests** for pure functions (dedup keys, cycle guard, seq decode, temporal boundary math).
+- **Integration tests** for multi-step sequences (crash-mid-write, corrupt-then-read, merge-into-then-read, contradicts-then-asOf) over a REAL `FileTreeMemoryStore` + `InMemoryCosineIndex` (mirror the L3 e2e fixtures — no stubbed store/index).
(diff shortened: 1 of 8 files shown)


ErikFortune:
@coderabbitai review

---

PR #519, opened 2026-07-07
feat(crypto-utils): base64url-no-pad helpers + branded MultibaseSpkiPublicKey

ErikFortune:
## Summary

Additive hardening of the **active** `@fgv/ts-extras/crypto-utils` surface for the PersonAIlity V2 identity/signing track (RFC 9421 + WebAuthn). Two coupled asks (base64url + multibase SPKI).

### Item 3 — bare base64url-no-pad primitives (extract + delegate)
- New `base64UrlNoPadEncode(data: Uint8Array): string` and `base64UrlNoPadDecode(encoded: string): Result<Uint8Array>` — RFC 4648 §5 base64url, no padding, **no** multibase `'m'` prefix.
- Extracted the body logic out of the existing `multibaseBase64UrlEncode/Decode`; those now **delegate** to the bare primitives (`multibaseBase64UrlEncode(d) === 'm' + base64UrlNoPadEncode(d)`; decode = `'m'`-prefix check → `base64UrlNoPadDecode(body)`). One implementation, not two. Existing multibase error messages/behavior are byte-for-byte preserved and the genuine browser-only `btoa`/`atob` `c8 ignore` branches are retained.

### Item 2 — branded `MultibaseSpkiPublicKey`
- Brand `MultibaseSpkiPublicKey = Brand<string, 'MultibaseSpkiPublicKey'>` (model.ts).
- Type guard `isValidMultibaseSpkiPublicKey` + validating `Converters.multibaseSpkiPublicKey`. Guard and decoder share **one** private shape rule (`isBase64UrlNoPadBody`) — no divergent regex.
- Tightened signatures (active surface): `exportPublicKeyAsMultibaseSpki → Promise<Result<MultibaseSpkiPublicKey>>` (brands the freshly-built output at the single construction site) and `importPublicKeyFromMultibaseSpki(encoded: MultibaseSpkiPublicKey, ...)`.
- New helpers/guard/type/converter exported from both `index.ts` and `index.browser.ts`.

### Item 5 — docs only
- `docs/FUTURE.md`: **"RFC 9421 HTTP-message-signature packlet (prospective @fgv/ts-extras)"** entry noting it depends on this stream's `base64UrlNoPadEncode/Decode`.

## api.md additive diff
```
+ function base64UrlNoPadEncode(data: Uint8Array): string;
+ function base64UrlNoPadDecode(encoded: string): Result<Uint8Array>;
[...]

PR #530, opened 2026-07-11
feat(crypto-utils): export MultibaseSpkiPublicKeyRegExp; widen import to accept string

ErikFortune:
## Summary

Two small, additive follow-ups to the `MultibaseSpkiPublicKey` surface in `@fgv/ts-extras` crypto-utils (the brand/guard/converter and the widened `exportPublicKeyAsMultibaseSpki` already shipped in #519). These close the last gaps so consumers can retire their local re-declarations and share one source of truth.

## Changes

1. **Export `MultibaseSpkiPublicKeyRegExp`** — the multibase-`m` + non-empty base64url-no-pad shape (`/^m[A-Za-z0-9_-]+$/`), co-located with the guard/converter. **It is load-bearing in `isValidMultibaseSpkiPublicKey`**: the guard now tests against this exact pattern and *additionally* enforces the base64url length rule shared with `base64UrlNoPadDecode`. So the exported pattern and the guard cannot silently diverge — the whole point of consolidating.

2. **Widen `importPublicKeyFromMultibaseSpki` to accept `string`** (was branded-only `MultibaseSpkiPublicKey`). A caller holding an unbranded value read from storage or the wire can now import it directly, without first routing it through the guard/converter. This is a **source-compatible widening** — a branded value is already a `string`, so every existing caller still compiles.

## Behavior note (intentional, safe)

Aligning the guard with the exported pattern tightens one degenerate case: `isValidMultibaseSpkiPublicKey('m')` (the `'m'` prefix with an **empty** body) now returns `false` (previously `true`). No real key ever encodes to an empty body — `exportPublicKeyAsMultibaseSpki` always produces a many-character base64url body — so **no valid key is affected**. Two existing tests that pinned the old empty-body acceptance were updated to assert the stricter behavior; the per-algorithm round-trip tests (all five `KeyPairAlgorithm`s) confirm real exported keys still validate and import.

## Validation posture (unchanged, documented)

The guard/regexp remain a structural **shape** check (prefix + base64url charset + valid base64 length), not full DER/SPKI/algorithm validation — a malformed-but-well-shaped string still fails later, with clear context, in `importPublicKeyFromMultibaseSpki`. The TSDoc states this explicitly on both the regexp and the guard. Result posture preserved: the converter returns `fail(...)`, never throws.

## Tests added

- `MultibaseSpkiPublicKeyRegExp` matches/rejects (incl. the empty-body, wrong-prefix, and non-base64url cases).
- Guard + converter reject the degenerate `'m'`.
- `importPublicKeyFromMultibaseSpki` accepts a plain (`string`-typed, unbranded) value and round-trips.

## Gates

`rushx build` ✅ · `rushx lint` ✅ (clean, `fixlint` run) · `rushx test` ✅ **100% coverage** (all four metrics). `api.md` updated for the new export + widened signature (committed). Changeset: `minor` (additive exports + source-compatible widening).

---
_Generated by [Claude Code](https://claude.ai/code/session_01T4rVgY8NFE7zNBNLgEVCch)_

## Summary by CodeRabbit

* **New Features**
  * Added a reusable pattern for validating multibase SPKI public keys.
  * Public-key import now accepts regular strings and reports malformed input through standard validation errors.

* **Bug Fixes**
  * Invalid keys with only the multibase prefix (`"m"`) are now correctly rejected.
  * Strengthened validation to require a non-empty, properly formatted key body.

### DIFF
--- a/common/changes/@fgv/ts-extras/multibase-spki-brand-followup_2026-07-08-17-30.json
+++ b/common/changes/@fgv/ts-extras/multibase-spki-brand-followup_2026-07-08-17-30.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "crypto-utils: export the MultibaseSpkiPublicKeyRegExp shape pattern next to the existing MultibaseSpkiPublicKey brand/guard/converter (single source of truth for consumers that need the pattern directly), and widen importPublicKeyFromMultibaseSpki to accept a plain string so callers holding an unbranded value can import it without first branding. The isValidMultibaseSpkiPublicKey guard now requires a non-empty base64url body (rejects the degenerate 'm'), aligning it with the exported pattern; no real exported key is affected.",
+      "type": "minor",
+      "packageName": "@fgv/ts-extras"
+    }
+  ],
+  "packageName": "@fgv/ts-extras",
+  "email": "noreply@anthropic.com"
+}
--- a/libraries/ts-extras/etc/ts-extras.api.md
+++ b/libraries/ts-extras/etc/ts-extras.api.md
@@ -459,6 +459,7 @@ declare namespace CryptoUtils {
         exportPublicKeyAsMultibaseSpki,
         importPublicKeyFromMultibaseSpki,
         isValidMultibaseSpkiPublicKey,
+        MultibaseSpkiPublicKeyRegExp,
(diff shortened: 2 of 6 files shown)


ErikFortune:
@coderabbitai review

---

PR #577, opened 2026-07-31
fix(ai-assist): guard capability resolvers against unresolved model aliases

ErikFortune:
## The defect

`AiAssist.resolveImageCapability` and its sibling `resolveEmbeddingCapability` prefix-match the supplied model id against the descriptor's capability rules (longest `modelPrefix` wins) with **no `MODEL_ALIAS_SIGIL` handling and no alias resolution**.

Because every provider declares a catch-all rule with `modelPrefix: ''`, an fgv model alias (`@<provider>:<role>`) matched the catch-all and the function returned a **confidently wrong capability** rather than failing or returning `undefined`. Reported by PersonAIlity; the sibling embedding resolver had the identical flaw.

| Provider | Alias passed in | Returned (before) | Correct (from concrete id) |
|---|---|---|---|
| `openai` | `@openai:image` | `{modelPrefix:'', outputParamStyle:'response-format'}` | `{modelPrefix:'gpt-image-', acceptsImageReferenceInput:true, outputParamStyle:'output-format'}` |
| `xai-grok` | `@xai-grok:imagine` | `{format:'xai-images', acceptsImageReferenceInput:false}` | `{format:'xai-images-edits', acceptsImageReferenceInput:true}` |
| `openai` (embedding) | `@openai:embedding` | `{modelPrefix:''}` | `{modelPrefix:'text-embedding-3', supportsDimensions:true, maxBatchSize:2048}` |
| `google-gemini` | `@google-gemini:flash-image` | (single catch-all — no mismatch) | same |

The xAI row flips the **wire format** and `acceptsImageReferenceInput`; the embedding row silently drops `supportsDimensions` and the `maxBatchSize` guard. A consumer branching on these builds a wrong request.

**The library's own paths were already safe** — `apiClient.ts` and `embeddingClient.ts` both call `resolveProviderModel` before the capability lookup. Verified and left untouched. The hazard was exclusively for direct consumers of the exported helpers, which is what they are exported for.

## Design choice — resolve the alias inside the helper

Two designs were on the table. **Chose (b): resolve the alias first inside the helper**, via a shared `@internal` generic `resolveCapabilityForModel<TCapability extends { readonly modelPrefix: string }>`.

| Input | Before | After |
|---|---|---|
| Concrete provider id | correct | **unchanged** |
| Registered fgv alias | catch-all → **wrong capability** | correct (identical to the concrete id's) |
| Unregistered / cyclic alias | catch-all → wrong capability | `undefined` |

**Why not (a) (`Result<T>` return).** The decisive reason is a hard constraint: (a) forces edits to `apiClient.ts`, which the stream brief explicitly forbids ("verify only, do not change"), and to `embeddingClient.ts`, outside the declared file surface. There is no way to land it inside scope. Independent of that, (b) is the smaller consumer-facing change and makes *both* input forms correct rather than turning one into an error.

**On `undefined` for an unresolvable alias.** The brief's objection was that `undefined` collides with "no rule matched". That objection targets the *wrong-capability* outcome, which (b) eliminates outright — a sigil-prefixed string now either resolves to a concrete id (→ correct capability) or yields `undefined`; it can never reach the catch-all. An unresolvable alias names no model, so "no capability applies" is truthful rather than lossy, and it fails in the safe direction. Both callers already surface `undefined` as a loud `Result.fail`. Note that pre-fix, `undefined` was effectively **unreachable** for any provider with a catch-all — that unreachability *was* the bug; post-fix it becomes a meaningful signal.

**Non-breaking:** public signatures unchanged. `etc/ts-extras.api.md` regenerates **byte-identical to `release`** — hence no api.md diff in this PR, and no merge conflict on that file from this side.

## Testbed fix

`samples/testbed/src/scenarios/imageGeneration/index.tsx`: `defaultModelFor` switched from the bare `AiAssist.resolveModel(descriptor.defaultModel, 'image')` spec walk (which yields the alias form) to the `AiAssist.resolveProviderModel` chokepoint (which yields the concrete id).

The value has a third consumer beyond the capability lookup: `SettingsPanel` renders it as a user-editable `<input value={model}>` directly beside a fetched dropdown of concrete ids (`gpt-image-1`, `dall-e-3`). Showing `@openai:image` there is incoherent, and any user edit replaces it with a concrete id anyway. Wire-path safety is unaffected (the library resolves either form). Secondary benefit: the scenario is now correct **independently** of the library guard, so a regression in either layer alone is still caught. The rewrite also replaced an imperative `if (!descriptor)` guard with a Result chain.

## TSDoc (adjacent ask A) — doc-only

`@remarks` on `ModelSpecKey` and an added paragraph on `resolveProviderModel` record that `tools` and `thinking` are deliberately **not** model selectors — they are orthogonal request params riding on top of whatever model the tier selected, removed from `ModelSpecKey` when the quality-tier axis landed. Both tell a tool-path caller to pass a tier and **not** hand-roll a `resolveModel` + `resolveModelAlias` walk. This rule previously lived only in `.ai/instructions/LIBRARY_CAPABILITIES.md`, and its absence from the API surface had sent two consumers down that unnecessary walk. **No `'tools'` key was added.**

## Test that fails against pre-fix behavior

Verified empirically by temporarily reverting the guard: **7 of the new tests fail**, while **all 62 pre-existing registry tests still pass** — demonstrating the catch and the no-regression-for-concrete-ids property in one run. Coverage includes all three reported mismatches, all three unresolvable-alias cases (unregistered, no-aliases-map, cyclic), an interior-`@` concrete id, and a registry-wide **invariant sweep** asserting that for every alias on every built-in provider, the alias and its concrete target select the identical capability for both modalities — which will catch a future provider that adds a specific-prefix rule alongside a new alias.

## Gates

| Gate | `@fgv/ts-extras` | `@fgv/testbed` |
|---|---|---|
| `rushx build` | pass (no api-extractor warning) | pass |
| `rushx lint` | pass | pass |
| `rushx test` | pass, 100%; `registry.ts` 100/100/100/100 | pass, 100%; `scenarios/imageGeneration` 100% |
| `rushx fixlint` | run before final commit | run before final commit |
| `rush change --verify` | pass | n/a |

No coverage-closure pass was needed — the scenario-driven tests reached 100% unaided, so **no `c8 ignore` directives were added**. `code-reviewer` was still run before declaring coverage done, per the load-bearing ordering in TESTING_GUIDELINES.

## Layer-1 review summary

`code-reviewer` run on the final diff: **Approved — zero findings at P1, P2, and P3.**

It independently verified the gates and the file-scope list, and ran its own sweep for other alias-leak sites, clearing two candidates: `applyCapabilityConfig` (`apiClient.ts:1527`, `idPattern`-based) is fed only concrete ids returned by a live list-models API, and `mergeThinkingConfig`'s `resolvedModel` comes from `resolveProviderModel` at both call sites. Its conclusion matched mine — the two functions fixed here were the only vulnerable lookup sites in the packlet. It specifically endorsed the shared-helper factoring (the duplicated-flaw history is the argument *for* consolidating), the `.orDefault()` + early-return shape over a fuller Result chain (the failure detail is deliberately discarded because no caller distinguishes unknown-alias from cyclic-alias), and confirmed the TSDoc claims match the implementation.

## Surfaced for the orchestrator — deliberately not fixed here

1. **Two sibling exported predicates share the same alias-leak class.** `isResponsesOnlyModel` and `isAdaptiveThinkingModel` (both `@public`, both in `model.ts`) also prefix-match a caller-supplied id with no sigil handling. Concretely: `isAdaptiveThinkingModel(anthropicDescriptor, '@anthropic:sonnet')` returns `false` where the concrete `'claude-sonnet-5'` returns `true`. **Not fixed** because `model.ts` is TSDoc-only in this stream's brief and CODING_STANDARDS asks for a brief amendment before extending outside the declared surface. Lower priority than the capability bug: both are called internally only with already-resolved ids, both already document "concrete (already-resolved) model id" as a precondition, and the failure mode is a loud provider-side HTTP 400 rather than a silently wrong request body. Recommend a small follow-on stream owning `model.ts`.

2. **The open P3 TECH_DEBT entry (`docs/TECH_DEBT.md:164`) stays open** — it asks for exactly design (a), with trigger "next substantive change to the capability resolution path". This stream is that trigger, but the file-surface constraint defers it. Not edited (`docs/` is outside scope). The guard shipped here removes the *dangerous* half, leaving only the non-idiomatic-return half.

3. `docs/WORKSTREAMS.md` not updated — orchestrator-owned.

Stream artifacts: `.ai/tasks/active/ai-assist-alias-capability-guard/{brief,state,result}.md`.

---

### DIFF
--- a/.ai/tasks/active/ai-assist-alias-capability-guard/brief.md
+++ b/.ai/tasks/active/ai-assist-alias-capability-guard/brief.md
@@ -1,0 +1,119 @@
+# Stream brief — `ai-assist-alias-capability-guard`
+
+Recorded verbatim from the orchestrator kickoff (2026-07-31).
+
+## Branch
+
+`ai-assist-alias-capability-guard`, branched from `origin/release`.
+
+> Brief stated base `b689c99ca`; at fetch time `origin/release` was `fbb08cd55`
+> (one commit ahead — a rush-shim lockfile repair, #575). Branched from
+> `origin/release` per the instruction; noted in `state.md`.
+
+## Context (verified by the orchestrator)
+
+PersonAIlity (a consumer) reported that `AiAssist.resolveImageCapability` matches raw
+model-id prefixes with no alias resolution and no `MODEL_ALIAS_SIGIL` guard. The
+orchestrator confirmed it by execution and found it is WORSE than reported, and that a
+sibling function has the same flaw.
+
+The exported capability resolvers live in
+`libraries/ts-extras/src/packlets/ai-assist/registry.ts`:
+
+- `resolveImageCapability(descriptor, modelId)` — filters `descriptor.imageGeneration` by
+  `modelId.startsWith(cap.modelPrefix)`, longest prefix wins.
+- `resolveEmbeddingCapability(descriptor, modelId)` — identical shape over
+  `descriptor.embedding`.
+
+Because every provider declares a catch-all rule with `modelPrefix: ''`, an unresolved fgv
+alias (a string starting with `MODEL_ALIAS_SIGIL`, i.e. `'@'`) matches the catch-all and
+returns a **confidently wrong capability** rather than failing or returning undefined.
+Reproduced on `release`:
+
+| Provider | Alias passed in | Capability returned | Correct capability (from concrete id) |
+|---|---|---|---|
+| `openai` | `@openai:image` | `{modelPrefix:'', outputParamStyle:'response-format'}` | `{modelPrefix:'gpt-image-', acceptsImageReferenceInput:true, outputParamStyle:'output-format'}` |
+| `xai-grok` | `@xai-grok:imagine` | `{format:'xai-images', acceptsImageReferenceInput:false}` | `{format:'xai-images-edits', acceptsImageReferenceInput:true}` |
+| `openai` (embedding) | `@openai:embedding` | `{modelPrefix:''}` | `{modelPrefix:'text-embedding-3', supportsDimensions:true, maxBatchSize:2048}` |
+| `google-gemini` | `@google-gemini:flash-image` | (no mismatch — single catch-all rule) | same |
+
+Note the xAI row flips the **wire format** and `acceptsImageReferenceInput`; the embedding
+row silently drops `supportsDimensions` and the `maxBatchSize` guard. A consumer branching
+on these builds a wrong request.
+
+**The library's own paths are already safe** — `apiClient.ts` resolves the alias via
+`resolveProviderModel` (~line 1336) BEFORE calling `resolveImageCapability` (~line 1341).
+Do not "fix" that; verify it and leave it. The hazard is exclusively for consumers calling
+the exported helpers directly, which is what they are exported for.
+
+**In-repo instance to fix:** `samples/testbed/src/scenarios/imageGeneration/index.tsx`.
+`defaultModelFor()` (~line 45) returns `AiAssist.resolveModel(descriptor.defaultModel, 'image')`,
+which yields the ALIAS form (e.g. `'@openai:image'`), and that value is fed to
+`AiAssist.resolveImageCapability` (~line 89). Result: wrong capability for the OpenAI and
+xAI defaults, which suppresses the "use as reference" affordance and offers wrong
+size/quality options.
+
+## Scope
+
+**1. Guard both capability resolvers (the core fix).** Decide between two designs and
+justify the choice:
+
+- (a) **Fail loudly** on a sigil-prefixed id — returning `undefined` is NOT acceptable
+  because `undefined` already means "no rule matched". Consider changing the return to
+  `Result<T>`, but that is a BREAKING signature change on an established export.
+- (b) **Resolve the alias first** inside the helper (it already receives the `descriptor`,
+  which carries `aliases`), so passing either form works.
+
+The orchestrator's lean is toward a design where an unresolved alias cannot silently
+produce a wrong answer. The chosen design must make the wrong-capability outcome
+impossible, and must not change behavior for concrete (non-alias) ids.
+
+**2. Fix the testbed call site** so the scenario resolves the correct capability for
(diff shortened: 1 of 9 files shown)


ErikFortune:
**Copilot loop: stopping at 2 rounds on diminishing returns.** Orchestrator drove rounds 1–2 after the implementing agent's session ended.

- **Round 1** — one substantive finding: the `ModelSpecKey` `@remarks` claimed "Every base model is thinking-capable", contradicted by five descriptors declaring `thinkingMode: 'unsupported'`. Real doc-accuracy defect, fixed in `fff87ca95`.
- **Round 2** — **no new comments**, one suppressed. The suppressed one was correct and is now fixed in `13b5fb411`: the round-1 rewording called thinking support "a per-provider/per-model property" and then pointed at `thinkingMode`, which is provider-level only. The descriptor doesn't encode per-model thinking availability at all. Reworded to say exactly that, and to state the tier-axis guarantee in its narrow, exact form — a tier selects a model *within* one provider and never changes the provider, so it never changes `thinkingMode`.

Round 2 surfacing zero new comments plus one suppressed doc nit is the diminishing-returns signal, so the loop stops here rather than commissioning a round 3.

**One note worth carrying forward.** The first attempt at the round-2 fix used `{@link IAiProviderDescriptor.thinkingMode}` and `{@link IAiProviderDescriptor.adaptiveThinkingModelPrefixes}`. `IAiProviderDescriptor` is not a package-root export, so api-extractor baked two `ae-unresolved-link` warnings straight into `etc/ts-extras.api.md` — the exact P3 in `CODE_REVIEW_CHECKLIST.md` § Documentation. Caught by rebuilding and diffing `api.md` before committing; reverted to code spans, and `api.md` is unchanged in the final commit.

Local gates on the final commit: `rushx build` ✅, `rushx lint` ✅, `rushx test` ✅ 100%. `api.md` byte-identical to `release`.

Erik has reviewed and cleared this to merge on green — merging once CI passes on `13b5fb411`.

---

PR #582, opened 2026-07-31
feat(ts-agent-memory): inject the derived record index into FileTreeMemoryStore.create()

ErikFortune:
## Summary

Adds one additive optional param to the public store factory:

```ts
FileTreeMemoryStore.create({ ..., index?: IMemoryIndex })
```

`FileTreeMemoryStore` already typed its index field as `IMemoryIndex` internally (`IInternalParams.index`, `this._index = params.index`), but the public `create()` hardcoded `MemoryIndex.create()` into those internal params, so a caller had no way to supply their own. That was the one genuine gap in what orchestrator triage found to be an already-mostly-present seam (`IMemoryIndex` is a real interface, `MemoryIndex implements` it, the retrievers already take it injected, `patch(op, entry)` already provides incremental maintenance, and `IMemoryEnvelope.contentHash` is already the staleness primitive).

- **Omitted** → a fresh `MemoryIndex`, byte-identical to prior behavior.
- **Supplied** → the store uses it for every index operation it performs and holds no second index: the `create()` `rebuild` of the initial vault walk, the `patch` on every write / delete / version invalidation / cap-cull eviction, and every `entries()` read behind `list` / `listScoped`, the keyed temporal reads, and the write path's content-hash dedup and write-policy admission cohort.

`IMemoryIndex`'s method signatures and return types are **unchanged** — the partial-read redesign is separate, design-first work and is explicitly out of scope here.

## Why — and the non-guarantee, which is load-bearing

This is **not** an experimentation seam for alternative index backends, and the TSDoc says so in a dedicated `@remarks` block opening with:

> **This is an instrumentation seam, NOT a resident-memory fix.**

`IMemoryIndex`'s read surface returns whole records by construction — `entries()` yields `IIndexedMemoryRecord`s and `byKind` / `byTag` / `byRecency` / `byRank` yield `IMemoryRecord<unknown>` = `{ envelope, body }` with the body materialized — so **any** index injected behind the current contract must still be able to produce every body. Injecting a "persisted" or "lazy" index today moves the storage and keeps the resident-memory ceiling. The doc's formulation: *an injected index changes WHERE records come from; it does not change WHETHER bodies are held.* Lowering the ceiling requires a separate, breaking, partial-read redesign of `IMemoryIndex` itself.

The intended use — stated immediately above the non-guarantee — is wrapping the shipped `MemoryIndex` in a counting/timing decorator to measure resident bytes by kind, open cost against vault size, and where the curve actually bends, so a later redesign decision is driven by numbers rather than estimates.

## Param name: `index`

Chosen over `memoryIndex` for two reasons:

1. It matches the existing private plumbing exactly — `IInternalParams.index` and `FileTreeMemoryStore._index` were already spelled `index`, so the public param is a pass-through with zero rename churn.
2. It matches the sibling convention in `IFileTreeMemoryStoreCreateParams`, which drops the domain prefix from the interface name (`registry: IBodyConverterRegistry`, `observers: IMemoryObserver[]`, `codecs: IIdentityCodec`). `IMemoryIndex` → `index` is the same transform. `vectorIndex` / `fragmentIndex` keep their qualifiers precisely because they are *not* the store's index — they are separate, differently-shaped seams that need disambiguating from it.

## Tests

New `src/test/unit/store/indexInjection.test.ts`, 8 scenarios.

**Default path (index omitted)** — serves `list` / `get` / `listScoped` from a default `MemoryIndex`; reopening an existing vault re-indexes it and resumes `seq` past the persisted high-water mark.

**Injected path**, via `RecordingMemoryIndex` — a faithful delegating decorator over the real `MemoryIndex` that records calls, deliberately the exact shape the consumer intends to build:

- receives the `create()` `rebuild` of an existing vault, exactly once, with both persisted records, and no `patch`;
- receives a `put` patch, a second `put` patch on a same-id revision, and a `delete` patch, each with the right `(op, scope, id)`;
- serves the store's read surface (`entries()` count rises across `list` / `listScoped`), and its projection views (`byKind` / `byTag` / `byRecency` / `byRank` / `backlinks` — none of which the store itself calls) answer from the patched state, proving no second index exists;
- a `RecencyRetriever` built over the *same* instance sees the store's records (the consumer's real pattern: store maintains, retrievers read).

Plus `HidingMemoryIndex` (filters `entries()`): the injected index — not a private one — determines what `list` / `listScoped` return, while `getById` still reads through to the FileTree, which remains the source of truth.

## Gates

| Gate | Status |
|---|---|
| `rushx build` | pass |
| `rushx lint` | pass (clean) |
| `rushx fixlint` before final commit | run; produced no changes |
| `rushx test` | 687 pass / 0 fail |
| Coverage | 100% statements / branches / functions / lines |
| No `any`; fallible ops return `Result<T>` | yes |
| `etc/ts-agent-memory.api.md` regenerated | yes (one added line) |
| Rush change file | yes (`minor`) |

Coverage reached 100% with no coverage-closure pass — no `c8` directives added, no coverage-driven tests written. Scenario tests were written first, per the required layer-1 ordering.

## Review

**No agent-spawn tool was available in the implementing session**, so the `code-reviewer` sub-agent could not be commissioned. Per the brief's process note I did not block on it and ran the review myself against `CODE_REVIEW_CHECKLIST.md` on the final diff.

- **P1 — none.** No `any`; no manual type-check-then-cast; no double casts; the one new fallible helper (`_resolveIndex`) returns `Result<IMemoryIndex>`; no `orThrow()` outside test setup.
- **P2 — one found and fixed (doc accuracy).** The first TSDoc draft enumerated the store's index consumers as "`rebuild`, `patch`, and every `entries()` read behind `list` / `listScoped` / the keyed temporal reads." Walking that back against the implementation showed the enumeration was incomplete in a *material* way: `_admissionCohort` (write-policy admission) and `_findByContentHash` (content-hash dedup) also read `entries()`, so an injected index that reshapes `entries()` changes **write** semantics, not merely read results. Fixed by restructuring into an explicit list of all three operation categories (naming dedup + admission), adding a caveat on why only a faithful delegating decorator is safe, and noting that `get` (flat) / `getById` go to the FileTree rather than the index. Test 8 was added to *evidence* the claim (an A/B showing content-hash dedup collapsing a same-body new id under the default index but not under the hiding index) rather than assert it in prose only.
- **P3 — none outstanding.** All `{@link}` targets are same-package exports, so no `ae-unresolved-link` warning is baked into `etc/*.api.md` — confirmed by a clean api-extractor run.

**Treat the above as a self-review, not an independent layer-1 pass.** If an independent `code-reviewer` pass is wanted on record, please commission it against this diff before merge.

Copilot review loop (layer 2) not yet driven.

## Notes for the orchestrator

- The concurrent vector stream will also regenerate `etc/ts-agent-memory.api.md`. This PR's change there is a single added line (`readonly index?: IMemoryIndex;` inside `IFileTreeMemoryStoreCreateParams`); resolve by rebuilding at integration, as planned.
- Nothing touched under `src/packlets/vector/**`, `libraries/ts-extras/**`, `samples/**`, `docs/WORKSTREAMS.md`, or `docs/STATUS.md`.
- Not merged, per the brief.

---

### DIFF
--- a/.ai/tasks/active/agent-memory-index-injection-seam/brief.md
+++ b/.ai/tasks/active/agent-memory-index-injection-seam/brief.md
@@ -1,0 +1,73 @@
+# Stream brief — `agent-memory-index-injection-seam`
+
+## Context
+
+The consumer (PersonAIlity) asked for "the `IVectorIndex` treatment" for the record index.
+Orchestrator triage found **most of it already exists**:
+
+- `IMemoryIndex` is a real interface (`libraries/ts-agent-memory/src/packlets/index/memoryIndex.ts`).
+- `MemoryIndex implements IMemoryIndex`.
+- `temporalRetrievers.ts` (and every other retriever) already consumes `IMemoryIndex` as an
+  injected dependency.
+- `patch(op, entry)` already provides incremental maintenance.
+- `IMemoryEnvelope.contentHash` is already the staleness primitive.
+
+The consumer has re-verified and accepted all of the above.
+
+**The one genuine gap:** `FileTreeMemoryStore` already types its index field as `IMemoryIndex`
+internally (`IInternalParams.index`, `this._index = params.index`), but the public `create()`
+hardcodes `MemoryIndex.create()` into those internal params. There is no way for a caller to
+supply their own.
+
+## Scope (deliberately narrow)
+
+Expose the index as an **optional injection point on the public `create()` params**, defaulting
+to the current concrete implementation. That is the whole deliverable.
+
+### Why the consumer wants it (this shapes the docs)
+
+**NOT** as an experimentation seam for alternative index backends. `IMemoryIndex`'s read surface
+returns full records by construction (`entries()` → `IIndexedMemoryRecord`;
+`byKind` / `byTag` / `byRecency` / `byRank` → `IMemoryRecord<unknown>` = `{envelope, body}`), so
+**any** index injected behind the current contract still materialises every body. Injecting a
+"persisted" index today moves the storage and keeps the resident-memory ceiling.
+
+It is wanted as an **instrumentation seam**: the consumer will wrap the shipped `MemoryIndex` in
+a counting/timing decorator to measure resident bytes by kind, open cost against vault size, and
+where the curve actually bends — so a later (breaking) partial-read redesign is driven by numbers
+rather than estimates. Their words: *"We've been asserting 'won't scale past thousands' without
+measuring it once."*
+
+**This distinction is load-bearing and must appear in the TSDoc.** A reader must not conclude
+that injecting a custom index fixes resident memory.
+
+## Requirements
+
+1. Optional index param on the public `FileTreeMemoryStore.create()` params, named consistently
+   with sibling optional params (`vectorIndex` / `fragmentIndex` / `observers`).
+2. Omitted → behaviour **byte-identical** to today (`MemoryIndex.create()`).
+3. Supplied → the store uses it for **every** index operation it currently performs, including
+   `rebuild` during `_initialIndex` and `patch` on every write.
+4. TSDoc per the "why" above, including the explicit non-guarantee.
+5. Tests: (a) default path unchanged; (b) an injected index receives the store's `rebuild` /
+   `patch` calls (decorator/spy over the real `MemoryIndex` — mirrors the consumer's intended
+   use); (c) an injected index's query results are the ones the store returns.
+
+## Out of scope
+
+- Any change to `IMemoryIndex`'s method signatures or return types (the partial-read redesign is
+  separate, design-first work).
+- Backlink type-discrimination (sibling stream).
+- `libraries/ts-agent-memory/src/packlets/vector/**` (concurrent stream owns it).
+- `libraries/ts-extras/**`, `samples/**`.
+- `docs/WORKSTREAMS.md`, `docs/STATUS.md` (orchestrator-owned).
+
+## Gates
+
+- `rushx build` / `rushx lint` / `rushx test` at 100% coverage in `libraries/ts-agent-memory`
+- `rushx fixlint` before the final commit
+- No `any`; fallible ops return `Result<T>`
+- Scenario tests BEFORE coverage-chasing; `code-reviewer` on the diff BEFORE the coverage-closure
+  pass
+- Rush change file; `etc/ts-agent-memory.api.md` regenerated
+- Commit, push, PR against `release`. Do NOT merge.
--- a/.ai/tasks/active/agent-memory-index-injection-seam/result.md
+++ b/.ai/tasks/active/agent-memory-index-injection-seam/result.md
@@ -1,0 +1,141 @@
+# Result — `agent-memory-index-injection-seam`
+
+## What shipped
+
+One additive optional param on the public store factory:
+
+```ts
+FileTreeMemoryStore.create({ ..., index?: IMemoryIndex })
+```
+
+- **Omitted** → `MemoryIndex.create()`, byte-identical to prior behavior.
+- **Supplied** → the store uses it for every index operation it performs and holds no
+  second index.
+
+Files changed:
+
+| File | Change |
+|---|---|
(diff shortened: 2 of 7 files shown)


ErikFortune:
## Independent layer-1 pass run and resolved (orchestrator)

The implementing agent's session had no agent-spawn tool and asked in the PR description for an independent `code-reviewer` pass on record before merge. That pass has now been run against `52cea8e0d`. Summary of its outcome and what was done about it.

### P1 — none

No `any`, no unsafe or double casts, no manual type-check-then-cast, no `Result<void>`, no `orThrow()` outside test setup. `_resolveIndex` is correctly typed and `Result`-returning.

### P2 — one finding, **fixed** in `c8695f438`

The reviewer confirmed the `@remarks` enumeration is now **accurate and complete** — it independently walked every `this._index` call site (`list`, `listScoped`, versioned `get`, `_persist`, `_deleteFlat`, `_evict`, `_admissionCohort`, `_findByContentHash`, `_versionsForEntity`, `_initialIndex`) against the doc's three categories and found no remaining omitted consumer.

But it flagged that the two claims added by the immediately-preceding fix commit were **asserted without being test-evidenced with an injected index**. The existing "reshapes `entries()` changes write semantics" test covered only the flat content-hash-dedup path. The temporal and cap-cull paths showed 100% measured coverage — but via pre-existing tests using the *default* index, which says nothing about whether an injected index actually drives them. That is the `TESTING_GUIDELINES.md` § "Coverage Gap Resolution" pattern exactly: the coverage tool measures the lines you have, not the test class you should have.

Fixed rather than dispositioned, since the claim in question is the one this PR most recently added and the most likely to silently drift. Two tests added:

1. **Versioned put + delete through a `HidingMemoryIndex`.** A hidden `v1` is invisible to the version walk, so the second put sees no current version, is built as a *first* version, and invalidates nothing; a subsequent soft delete tombstones only what the index shows as current. `v1` remains on disk carrying no `invalid_at` — two live currents, which is precisely the documented hazard.
2. **Cap-cull through both decorators.** `RecordingMemoryIndex` proves the eviction reaches the *injected* index as a `delete` patch (so no private index was patched instead); `HidingMemoryIndex` proves that hiding a cohort member shrinks the admission cohort under the cap, so nothing is evicted and the record survives.

Both discriminate rather than pass vacuously: if `_versionsForEntity` or `_admissionCohort` stopped reading through the injected index, the hiding cases would match their controls and these tests fail.

### P3 — one advisory, not actioned

The default-index "reopening resumes seq" test overlaps `fileTreeMemoryStore.test.ts`. It reads as a deliberate baseline for the injected-index rebuild test that follows it, so leaving it.

### Gates on `c8695f438`

`rushx build` ✅ · `rushx lint` ✅ (clean) · `rushx test` ✅ 687→689 passing, `indexInjection.test` 10/10.

### Copilot loop

Rounds 1 and 2 both returned **no comments** (round 2 ran against the doc fix). Stopping there on diminishing returns. Requesting one more round now that the tests changed, since that is a real diff rather than a doc-only edit.

---

ErikFortune:
**Copilot loop: stopping at 3 rounds on diminishing returns.**

Rounds 1, 2, and 3 all returned **no new comments**. Round 3 (requested because the test-evidence commit was a real diff, not a doc-only edit) carried one suppressed advisory, which turned out to be correct and is fixed in `3b41555ee`.

`_resolveIndex` tested `index !== undefined`, while **all eight** sibling optional params in `create()` absorb absence with `??` — which takes `null` and `undefined` alike (`params.codecs ?? new Map()`, `params.clock ?? Date.now`, and so on). Verified that inconsistency directly rather than taking the report on faith. A JS caller, or a TS caller arriving through an `unknown` escape hatch, passing `null` got `null` installed as the store's index and failed later inside `entries()` with a message naming neither the param nor the cause.

Fixed with a nullish guard plus a test asserting a `null` index yields a working store on the default index. This is the same class as #583's `classifyJsonParseFailure` guard — a boundary that is total for a well-typed caller and not for anyone else — which is why it was worth fixing rather than dispositioning despite being suppressed.

Three consecutive rounds of zero surfaced comments is a clear diminishing-returns signal, so no round 4.

**Gates on `3b41555ee`:** `rushx build` ✅ · `rushx lint` ✅ (clean) · `rushx test` ✅ 690 passing, `indexInjection.test` 11/11 · `api.md` unchanged by this commit.

### State for merge

Layer 1 (independent `code-reviewer`) run and its P2 resolved; layer 2 closed at 3 rounds. Awaiting your review — and please merge **this before #585**, since both regenerate `etc/ts-agent-memory.api.md` and this one's change there is a single added line.

---

PR #583, opened 2026-07-31
fix(ai-assist): make classifyJsonParseFailure total for non-string input

ErikFortune:
Follow-up to #579, from that PR's Copilot round-1 finding.

## The defect

`classifyJsonParseFailure` is documented as never failing — it returns a `JsonParseFailureReason` rather than a `Result`, so a caller has no failure channel to check. But unlike its sibling `extractJsonText`, it had no runtime guard for non-string input, and `locateJsonCandidate` throws on one.

Verified empirically against the built library before fixing:

```
classifyJsonParseFailure(42)  -> THROWS: TypeError - text.replace is not a function
extractJsonText(42)           -> Result failure: extractJsonText: input must be a string.
```

So the doc claim and the implementation contradicted each other, and the two sibling entry points disagreed on how they handle the same bad input.

The signature says `string`, so a well-typed TS caller can't hit this. A JS consumer, or a TS caller arriving through an `unknown`/`any` escape hatch (which is exactly the shape of the LLM-response paths this helper serves), can.

## The fix

A `typeof text !== 'string'` guard returning `UNKNOWN_PARSE_FAILURE` — the same disposition the function already uses for input it can't classify. This makes "never fails" true in practice rather than only under a well-typed caller, and makes the guard behavior consistent with `extractJsonText`.

Also adds a test covering `42`, `null`, `undefined`, `{}`, `[]`, and `true` — asserting both that no throw escapes and that the result is `UNKNOWN`.

## Gates

- `rushx build` — pass
- `rushx lint` — pass (`rushx fixlint` run before the commit)
- `rushx test` — pass, **100% coverage on all files** in `ts-extras`
- No `any`; the test's deliberate bad input goes through `as unknown as string`

## Review

Layer 1 (`code-reviewer`) is not separately run here — this PR *is* the resolution of a layer-2 finding on the already-merged #579, and the change is a three-line guard plus its test. Copilot will be requested once CI is green.

Change file: `common/changes/@fgv/ts-extras/fix-classify-json-non-string-guard_*.json` (`patch`).

---


ErikFortune:
**Copilot loop: stopping at 1 round.** Approved with 0 comments generated across all 3 changed files. CI green.

One round is the right stop here rather than a premature call: the diff is a three-line `typeof` guard plus its test, the defect was itself surfaced by Copilot on #579, and the fix was verified empirically against the built library before being written (`classifyJsonParseFailure(42)` threw `TypeError: text.replace is not a function`; `extractJsonText(42)` returned a clean `Result` failure). There is no surface here for a second round to find anything on.

**Not merging without your review** — your merge authorization covered #576–#579 specifically, and this is a new PR. It should be a fast read: `libraries/ts-extras/src/packlets/ai-assist/jsonResponse.ts` is the only production change.

---

PR #585, opened 2026-07-31
feat(agent-memory): durable fragment identity — opaque `fragmentId` + advisory `locator`

ErikFortune:
## Why

A consumer is moving knowledge ingestion to **model-chosen segmentation**, where the model may *rewrite* a span into a curated block rather than only choose boundaries. Two consequences: a fragment is not necessarily a substring of the body, and fragmentation is no longer re-derivable from the body. A fragment therefore needs a **durable identity of its own**.

Design settled upstream over two rounds; this PR implements it.

## `@fgv/ts-agent-memory`

- **`IEmbeddedFragment.fragmentId?: string`** — an **opaque bytestring**. Never parsed, never filtered on, no semantics, not in the query path. Carried back verbatim on **`IVectorQueryHit.fragmentId?`**.
- **`IEmbeddedFragment.locator` is now optional.** A fragment must carry **at least one** of `locator` / `fragmentId`.
- The at-least-one invariant is enforced in a **converter, not the type** — new `embeddedFragmentConverter` / `fragmentLocatorConverter` (`packlets/vector/fragmentConverters.ts`), re-checked at both index seams. The conditional-required union (`{locator; fragmentId?} | {locator?; fragmentId}`) was considered and declined: it costs at every construction site and buys nothing at the read site, where each field reads as `… | undefined` either way.
- `InMemoryFragmentCosineIndex` carries `fragmentId` through add → query.

## Documentation (the load-bearing half)

- **`IFragmentLocator`** now states the span is **advisory** — the region of the body a fragment was *derived from*, **NOT** a slice guaranteed to reproduce the fragment's text. `body.slice(start, end)` does not round-trip under a rewriting segmenter.
- **Struck the false claim** that `IFragmentVectorIndex` reuses `IVectorQueryHit` *"(whose `locator` is always populated here)"* — false the moment input `locator` became optional.
- **Rewrote `IVectorQueryHit.locator`**, which actively invited presence-branching. The new note:
  - states that **no *single* field discriminates** a fragment hit from a record hit (not the flat, and false, "neither field is a discriminant");
  - explicitly declines to offer the *disjunction* as a discriminator — it would couple callers to an invariant enforced at a different boundary, and fail silently if that invariant were relaxed;
  - gives the robust rule: **fragment-ness is determined by which index produced the hit** (`IFragmentVectorIndex.query` vs `IVectorIndex.query`);
  - notes `locator` absence now carries **two** meanings (record hit, or a fragment with no honest span) — exactly why presence-branching is unsafe.

## `@fgv/ts-agent-memory-sqlite-vec`

- Third auxiliary column **`+fragment_id text`** alongside `+start_off` / `+end_off`; carried on upsert, returned on the hit, never filtered.
- **Absent locator → an all-or-nothing NULL offset pair.** Write-side safe-integer validation now runs only when a locator is present (nothing to validate otherwise, and it must not reach `BigInt(undefined)`); read-side `_toOffset` coercion is unchanged. A **half-NULL pair** — or a row carrying **neither identity** — now fails the query loudly as corrupt persisted data rather than fabricating a span (`Number(null)` is `0`) or returning an anonymous hit.
- **Schema-migration detection (the required piece).** `CREATE VIRTUAL TABLE IF NOT EXISTS` is a no-op against an existing table — SQLite never compares schemas — and `vec0` has no `ALTER TABLE ADD COLUMN`. An existing 4-column fragment DB would therefore meet the widened 5-column INSERT as an opaque `no such column: fragment_id` at statement-prepare time. `create()` now reuses the **same stored-DDL parse that already recovers the dimension** to compare the auxiliary-column set, failing with an actionable message naming expected-vs-found columns and stating that a drop-and-re-index is required.

## Policy documentation

The **drop-and-re-index rule** is written down where a consumer will actually meet it — they are by definition upgrading, not reading a design thread:

- `.ai/instructions/LIBRARY_CAPABILITIES.md` (the `@fgv/ts-agent-memory-sqlite-vec` entry)
- the package README (a dedicated `## Upgrading` section)

Both state plainly: a `vec0` schema change requires consumers to drop the table and re-index; there are **no in-place migrations**; vectors are re-derivable from records, so it costs embedding time, **never data**.

## Decisions recorded — decided, not overlooked

Captured in `.ai/tasks/active/agent-memory-fragment-id/result.md` so a future consumer reopens them as **new asks** rather than as oversights:

- **No `exact` / `verbatim` flag on the locator.** Considered and declined: once fragments are durable state, span-to-text reproduction stops being an operation the consumer performs.
- **No conditional-required union** for the at-least-one invariant (see above).
- **No new fragment-vs-record discriminant field** on `IVectorQueryHit` — not needed; you query the index you chose.
- **Fragment-id stability across re-embeds is the consumer's responsibility.** The library's guarantee is "we never parse it", **not** "we keep it stable" — whole-record-replace means an updated record re-emits its whole fragment set. Documented as such on the field.

## Behaviour change beyond pure addition (disclosed)

`SqliteVecFragmentIndex.create` now **rejects** a pre-existing table of the configured name whose auxiliary columns do not match — including a plain non-`vec0` table, and a table with matching auxiliary columns but no `vec0` embedding column. Previously it succeeded with an unrecovered dimension and failed later at insert time. Two existing tests asserting the old late-failure shape were rewritten to assert the new create-time rejection: the detection moved earlier and the message became actionable; **nothing that previously worked now fails.**

Everything else is a pure addition. A hit for a fragment stored without a `fragmentId` **omits the key entirely** (never present-and-`undefined`), so existing structural comparisons are unaffected — asserted in both packages with `Object.keys(hit).sort()` + `toStrictEqual`.

## Tests

100% coverage in both packages. Highlights:

- `fragmentConverters.test.ts` (new) — the at-least-one invariant across all four identity shapes, plus field validation.
- In-memory index — opaque `fragmentId` round-trips verbatim; id-only, locator-only, both, and mixed-shape records; byte-identical hit shape when `fragmentId` is absent; neither-identity rejection.
- SQLite index — all of the above against real SQLite, plus `fragment_id` + NULL-locator persistence across a genuine close+reopen; safe-integer check skipped for a locator-less fragment and still enforced when a locator is supplied; **auxiliary-column schema detection** (actionable message naming expected-vs-found; asserts it states drop-and-re-index + no data loss; asserts the opaque `no such column` error is *not* what surfaces; order-insensitive acceptance; same-arity-different-name rejection; no-embedding-column rejection); and the two new corrupt-persisted-data guards.

## Gates

- [x] `rushx build` passes in both packages
- [x] `rushx lint` passes in both packages
- [x] `rushx test` passes at **100%** coverage in both packages
- [x] `rushx fixlint` run before the final commit
- [x] No `any`; all fallible operations return `Result<T>`
- [x] Rush change files for both packages
- [x] `etc/*.api.md` regenerated

## Review-loop discipline (layer 1)

**The `code-reviewer` sub-agent could not be run — this environment exposes no agent-spawn tool.** In its place I ran a manual pass against `CODE_REVIEW_CHECKLIST.md` on the **pre-coverage-closure** diff, per the required ordering. It produced three changes:

1. **P2** — replaced three `{@link AUXILIARY_COLUMNS}` references (a module-private const) with plain code spans; they were baking `ae-unresolved-link` warnings verbatim into the checked-in `etc/*.api.md`, the exact stale-api.md liability the checklist's TSDoc section names.
2. **P2** — asked "should this branch exist?" of the `match === null ? undefined : …` dimension fallback and concluded it should not: with the auxiliary-column check in place, a table that passes it but has no `float[<n>]` column is not a usable fragment index, and returning a dimensionless index there produced a half-state whose first add would `CREATE VIRTUAL TABLE IF NOT EXISTS` into a no-op. Replaced with a loud failure — which **removed** a would-be coverage gap rather than papering over it with a directive.
3. **P3** — documented why the shared `/g` `AUXILIARY_COLUMN_RE` is safe under `matchAll` (it iterates a clone; it does not advance the shared instance's `lastIndex`) — a footgun this repo has been bitten by before.

No `c8 ignore` directives were added. The two in-`captureResult` throws are corrupt-persisted-data guards following the file's established convention (`_parseKey`, `_toOffset`).

Note this is a **native-boundary package** (`better-sqlite3` / `sqlite-vec`), a known layer-1 blind spot — a substantive Copilot loop on rounds 1–2 is expected here and is *not* evidence layer 1 was skipped.

## Coordination

`libraries/ts-agent-memory/etc/ts-agent-memory.api.md` is regenerated here and will conflict with the concurrent `agent-memory-index-injection-seam` stream, which also regenerates it — **expected, resolve at integration**.

Out of scope per brief and untouched: `packlets/store/fileTreeMemoryStore.ts` create-params, the `index` packlet, `docs/WORKSTREAMS.md`, `docs/STATUS.md`. No store change was needed — `FragmentEmbedder` is consumer-supplied and returns `IEmbeddedFragment[]`, so `fragmentId` flows through the existing fragment-embed-on-write hook untouched.

---


ErikFortune:
## Independent layer-1 pass run — **approved clean** (orchestrator)

The implementing agent's session had no agent-spawn tool and ran a manual `CODE_REVIEW_CHECKLIST.md` pass instead. An independent `code-reviewer` pass has now been run against the head commit, briefed specifically on the native-boundary blind spot (`CODING_STANDARDS.md` § "Native-boundary packages…" — #562 on this same package had a clean layer-1 followed by three real runtime-mode findings from Copilot).

**Result: no P1s, no P2s, no new P3s.** The reviewer re-derived the conclusions from source rather than accepting the agent's report, and independently re-ran the gates: both packages build and lint clean; `ts-agent-memory` at 100% across all metrics; `ts-agent-memory-sqlite-vec` 71/71 with 100% on both index files.

The four items it was asked to scrutinize all check out against the actual persistence and query paths, not just stated intent:

1. **All-or-nothing NULL offset pair** — confirmed there is exactly *one* write path (`replaceTxn`, whole-record delete-then-insert in a single transaction; no separate upsert/update), and both offset binds sit behind the same `locator === undefined` check, so a half-NULL pair is unreachable by construction. The read side throws on one rather than coercing (`Number(null) === 0` would fabricate a span at the top of the body), and that throw is inside `query()`'s `captureResult`, so it surfaces as `Result.fail` rather than escaping. Two tests reproduce start-only and end-only corruption.
2. **Safe-integer guard narrowed, not weakened** — never reaches `BigInt(undefined)`, still fires unconditionally whenever a locator *is* supplied.
3. **Neither-identity row** — enforced on the **write** side in both index implementations before any mutation, plus the converter boundary, plus a loud read-side detector. Three independent enforcement points, not just read-side detection.
4. **Schema-change behavior** — matches the re-index policy point for point: detected on open via the same stored-DDL parse already used for dimension recovery, failing with expected-vs-found columns named and the remedy stated. The two rewritten tests reflect a genuine improvement (previously: succeeded with an unrecovered dimension, then failed later at insert with an opaque `no such column`). Nothing that previously worked now fails.

It also walked the discriminator TSDoc claim-by-claim and confirmed each against the implementation, including that the "no single field" phrasing is correctly scoped — the *pair* does discriminate, and the doc never claims otherwise; it only declines to expose that disjunction as a named API.

All `{@link}` targets resolve within their own package; the cross-package references are correctly plain code spans.

**Merge ordering note.** As flagged, `etc/ts-agent-memory.api.md` will conflict with #582. Plan is to merge **#582 first** (its api.md change is a single added line), then rebase this branch onto `release` and regenerate the report — it is generated, so regenerate rather than resolving the conflict textually.

Copilot review requested; will drive that loop and report the stop reason.

---

PR #596, opened 2026-08-02
feat(ts-utils): the async Result family

ErikFortune:
Implements `.claude/project/async-result-family-design.md` (#595) per `.ai/tasks/active/async-result-family/brief.md`. All five open questions in § 4 of the design were resolved before implementation and were implemented as resolved; none were relitigated.

## What shipped

One new file, `libraries/ts-utils/src/packlets/base/mapResultsAsync.ts`, re-exported through the existing `export * from './packlets/base'`.

**Five bounded-parallel collectors** — one scheduler, five folds, each mirroring its sync sibling's name, parameter order and fold:

| Async | Sync sibling | Returns |
|---|---|---|
| `mapResultsAsync` | `mapResults` | `Promise<Result<T[]>>` |
| `mapDetailedResultsAsync` | `mapDetailedResults` | `Promise<Result<T[]>>` |
| `mapSuccessAsync` | `mapSuccess` | `Promise<Result<T[]>>` |
| `mapFailuresAsync` | `mapFailures` | `Promise<string[]>` |
| `allSucceedAsync` | `allSucceed` | `Promise<Result<T>>` |

Each carries both forms (OQ-3): the thunk form `f(work, …)` and the `(items, fn, …)` form. Each takes `IAsyncResultOptions` (`concurrency?`, `aggregatedErrors?`).

**Two transitively-async members**, serial by contract, taking **no** options bag (§ 2c): `populateObjectAsync`, `firstSuccessAsync`. Both take an optional `IMessageAggregator` and nothing else — a `concurrency` field on a serial-by-definition operation would be a lie in the type signature.

**Plus** `AsyncDeferredResult<T>`, `AsyncFieldInitializers<T>`, `IAsyncResultOptions`, and `DEFAULT_RESULT_CONCURRENCY` (= 8, finite and exported so callers can read it rather than guess).

Everything is `@public` (OQ-5). No new `ae-incompatible-release-tags` and no new `ae-unresolved-link` warnings in `etc/ts-utils.api.md` — verified by diffing the regenerated file (the added-lines warning count is zero).

## The four things most likely to ship subtly wrong, and how each is pinned

- **Input order, not completion order.** `_scheduleBounded` writes into a pre-sized array at the input index. Pinned by a test that drives 30 ms / 15 ms / 1 ms work and asserts *both* that the observed completion order is the reverse of input order *and* that the results are in input order.
- **The bound actually binds.** The work functions are instrumented with an in-flight counter; tests assert the observed **peak** never exceeds `concurrency`. Nothing is inferred from timing. Covered for `concurrency: 3` over 20 operations, `concurrency: 1` (peak exactly 1), the default (peak ≤ `DEFAULT_RESULT_CONCURRENCY` and > 1, plus assertions that the default is finite and positive), fractional (truncated), infinite (peak = N), and `concurrency > work.length` (never starts more workers than there is work). `0` / negative / `NaN` are clamped to 1 — the `NaN` case matters because an unguarded `Math.min(NaN, …)` would have silently scheduled nothing and returned an array of holes.
- **A rejected promise is a `Failure`, not a throw.** Every unit of work goes through `captureAsyncResult`, which catches both a synchronous throw from the thunk (before it returns its promise) and a rejection of the returned promise. Both cases are tested separately on `mapResultsAsync`, and also on `mapFailuresAsync`, `allSucceedAsync`, `populateObjectAsync`, and `firstSuccessAsync` (which has both a sync-throw and a reject test). This is the class of defect #594 caught in the safer-fetch core.
- **`map*` keeps scheduling after a failure; `allSucceedAsync` does not short-circuit.** Both look like missed optimizations, so both have a test that says otherwise: the work functions record completion, and the test asserts every unit ran *and* that every error was aggregated.

Empty input succeeds with an empty array for all five, matching the sync siblings; `firstSuccessAsync` on empty input fails with `no results found`, matching `firstSuccess`. `aggregatedErrors` threading matches the sync family per-member (aggregate-on-failure for `mapResults`/`mapSuccess`/`allSucceed`/`populateObject`, aggregate-always for `mapFailures`).

## Where the design was underspecified, and what I decided

Three places the code hit something the design did not settle. All are additive and none contradict a resolved OQ, but they're the things worth a reviewer's eye:

1. **`mapDetailedResultsAsync` and a rejected work function (deliberate, documented divergence from the sync fold).** The sync `mapDetailedResults` ignores any failure whose `detail` is falsy — its fold is `result.detail && !ignore.includes(result.detail)`. A thunk that throws or rejects produces *no detail at all*, so under a literal transposition of that fold, a rejected promise would be **silently swallowed**. That is clearly wrong, so a capture failure is treated as always-reported and never ignorable, and the divergence is called out in the TSDoc. The inner-`DetailedResult` fold is otherwise byte-identical to the sync one, falsy-detail quirk included (fixing that quirk is a sync-side change and out of scope here). Tested: an ignorable detail is ignored, a non-ignorable detail is reported, and a thrown work function is reported **even when `ignore` lists every detail value**.

2. **Overload dispatch is `typeof args[1] === 'function'`, which is unambiguous for four of five collectors but not for `allSucceedAsync`.** For `mapResultsAsync` / `mapSuccessAsync` / `mapFailuresAsync` the second argument is either `fn` or an options object; for `mapDetailedResultsAsync` it is either `fn` or the `ignore` array. Neither an options object nor an array is a function, so the discrimination is total. `allSucceedAsync`'s second argument is either `fn` or `successValue: T` — and `T` can itself be a function type. The design (OQ-3) anticipated this class and judged it rare, but noted it would surface as a type error; here it would instead be silent misdispatch, so dispatch additionally requires `args.length >= 3`. That makes `allSucceedAsync(work, someFunction)` correct, leaves only `allSucceedAsync(work, someFunction, options)` misread, and the residual case is documented in the TSDoc `@remarks`. There is a test pinning the function-valued-`successValue` behaviour.

3. **`concurrency` validation is clamping, not failure.** The design says the default must be finite but doesn't say what an invalid explicit value does. Failing would have been awkward — `mapFailuresAsync` returns `Promise<string[]>`, with no `Result` to fail into — so out-of-range values clamp to 1 and fractional values truncate. `Number.POSITIVE_INFINITY` falls out naturally as "unbounded", which seems like the right escape hatch for a caller who explicitly wants the old behaviour. Documented on `IAsyncResultOptions.concurrency`.

One smaller note: the four `Result`-based collectors delegate their folds to the *actual sync functions* (`mapResults`, `mapSuccess`, `mapFailures`, `allSucceed`) after flattening the scheduled outcomes into `Result<T>[]`. That was deliberate — it makes fold parity structural rather than a copy that can drift. Only `mapDetailedResultsAsync` writes its own loop, for reason (1) above.

## Explicitly not in scope

Cancellation / `AbortSignal` (OQ-4 — deferred; `AsyncDeferredResult` is **not** pre-widened defensively, since the design verified the retrofit is non-breaking); retry / timeout / backoff in the scheduler; a first-to-settle variant; migrating the in-repo hand-rolled call sites (`HybridRetriever`, the prompt-assist observer fan-out, the bcp47 registry loader).

## Docs

`.ai/instructions/LIBRARY_CAPABILITIES.md` updated in this PR, per the brief:
- The `@fgv/ts-utils` `base` packlet row now covers the whole async family, the two-way split, both call forms, and the in-input-order / no-short-circuit / rejection-becomes-`Failure` contracts.
- Two new "Decision shortcuts" entries: one for *"N independent async `Result`-returning operations with a concurrency bound"* (which states outright that you **cannot** pass `Promise<Result<T>>[]` and why — that misunderstanding is what produced this design — and points at `await Promise.all(xs.map(f))` + `mapResults` as the longhand to stop writing), and one for the serial-by-contract members so `populateObjectAsync` / `firstSuccessAsync` are findable without reading the collector entry first.

## Gates

- `rushx build` — passes; `etc/ts-utils.api.md` regenerated and committed.
- `rushx lint` — passes; `rushx fixlint` run before the final commit.
- `rushx test` — passes. 65 new tests; `mapResultsAsync.ts` is at **100% statements / branches / functions / lines**, and the package's global coverage gate passes.
- `rush build` (repo-wide) — run to confirm the additive exports don't disturb downstream consumers.
- Rush change file added (`minor` — new public API surface).

## Review-loop disclosure

**No agent-spawn tool was available in this session, so the `code-reviewer` agent could not be run.** Per the brief's contingency, I did a **manual pass against `.ai/instructions/CODE_REVIEW_CHECKLIST.md` instead** — and, per `TESTING_GUIDELINES.md` § "Coverage Gap Resolution", the sequence was still scenario-driven tests first, review second, coverage last. Worth noting that the coverage-closure pass turned out to be a no-op: the scenario-driven tests hit 100% on their own, so there are **no `c8 ignore` directives and no coverage-chasing artifacts** in this diff.

Manual pass findings:

- **P1 — none.** No `any` (verified by grep over both new files); no double casts; no manual-type-check-then-cast; no `orThrow`; nothing throws out of business logic. The only `as unknown as T` casts are two in the test file for intentionally-invalid test data (a `undefined` initializer and an `undefined` value for a `string` field), which `CODE_REVIEW_CHECKLIST.md` § "Special Cases / Test Files" sanctions.
- **P2 — two, both resolved.** (a) The `as` casts inherent to resolving an overload implementation were scattered across call sites in my first draft; they're now contained in a single documented private helper, `_asWorkArray`. (b) `IAsyncResultOptions.concurrency`'s TSDoc originally documented only the "below 1" clamp and not `NaN` or truncation, which is exactly the doc-drift class the checklist calls out — now documents all three, matching the tests.
- **P3 — one, resolved.** A module-private type alias was named `IAsyncWork`, where the `I` prefix wrongly implies an interface; renamed to `AsyncWork`.
- **Dispositioned, not fixed:** `mapResultsAsync`, `mapSuccessAsync` and `mapFailuresAsync` have near-identical four-line bodies. Extracting them would need a higher-order helper generic over the fold *and* the return type (one of the three returns `string[]`, not `Result<T[]>`), which reads worse than the repetition. Per `CODING_STANDARDS.md` § "Avoid Over-Engineering" — three similar lines beat a premature abstraction.

An independent `code-reviewer` pass may still be wanted before merge; the three underspecified-design decisions above are the places I'd point it at first.

Copilot review loop to follow; the stop reason will be recorded on this PR.

---


ErikFortune:
## Review-loop stop record

**Copilot review loop driven by implementer; stopped at 1 round on diminishing returns.**

Round 1 reviewed all 7 changed files and **generated no comments** ([review](https://github.com/ErikFortune/fgv/pull/596#pullrequestreview-4837580307)). Per `CODING_STANDARDS.md` § "Round count is not the signal — substantive value per round is", the stop criterion is the finding profile of the most recent round, not the count: a round that surfaces nothing gives a round 2 nothing to build on, so requesting another would be pure loop cost. Stopping at 1.

Two caveats worth flagging rather than burying, since a clean round 1 is the case where a stop record can be over-read:

1. **This is not evidence that layer 1 was thorough.** Layer 1 here was a *manual* pass against `CODE_REVIEW_CHECKLIST.md`, not the `code-reviewer` agent — no agent-spawn tool was available in the session. A zero-finding Copilot round does not substitute for the internal reviewer, which reads against *intent* rather than against the diff. An independent `code-reviewer` pass may still be worth commissioning before merge.
2. **The three design-underspecified decisions in the PR description are where I'd point that pass first** — particularly (1), `mapDetailedResultsAsync`'s deliberate divergence from the sync fold for capture failures. That is a judgement call about a swallowed-rejection hazard in the sync sibling's `result.detail && …` fold, and it is exactly the class of thing an automated reviewer reading the diff in isolation would not question, because the code does what its own TSDoc says.

### Gate status at this commit (`c9dfe3d`)

| Gate | Result |
|---|---|
| CI `build` | ✅ success |
| `rushx build` / `lint` / `test` (ts-utils) | ✅ pass; `rushx fixlint` run before final commit |
| Coverage, `mapResultsAsync.ts` | ✅ 100% statements / branches / functions / lines |
| `rush build` (repo-wide, local) | ✅ pass, 4m13s — additive exports disturb no downstream consumer |
| `etc/ts-utils.api.md` | ✅ regenerated; **zero** new `ae-incompatible-release-tags` and zero new `ae-unresolved-link` warnings on added lines |
| Rush change file | ✅ present (`minor`) |
| `LIBRARY_CAPABILITIES.md` | ✅ updated in this PR (base packlet row + 2 decision shortcuts) |

No `c8 ignore` directives were added — the scenario-driven tests reached 100% on their own, so the coverage-closure pass was a no-op.

---

PR #598, opened 2026-08-04
docs: correct design-doc status lines and record implementation divergences

ErikFortune:
Both design docs still announced themselves as unimplemented after their streams landed. That's the doc-accuracy class `CODE_REVIEW_CHECKLIST.md` calls out — *document what ships, not design intent* — and it's the kind of drift that misleads the next reader most cheaply.

Docs only.

## `async-result-family-design.md`

Marked **implemented and merged (#596)**, and retained as the rationale record rather than a statement of pending work.

New **§ 6 — divergences found in implementation**, so the doc describes what shipped:

- **`mapDetailedResultsAsync` deliberately diverges from its sync fold.** `mapDetailedResults` ignores a failure whose `detail` is falsy, and a rejected thunk has *no* detail — so a literal transposition would have **silently swallowed rejected promises**, the same class as the safer-fetch body-read escape. Capture failures are now always reported and never ignorable.
- **OQ-3's ambiguity assessment was too optimistic for one member.** I wrote that the ambiguous overload case "surfaces as a type error rather than silent wrong behaviour." True for four of five collectors; **not** for `allSucceedAsync`, whose second parameter is a caller-supplied `successValue: T`. A function-valued `successValue` plus options type-checked cleanly and then failed at runtime. Settled by discriminating on the *first* argument. The doc also records the **rejected** alternative — keying on `fn.length === 2` would misdispatch the common `(item) => ...` callback that ignores its index, trading a rare bug for a frequent one.
- **Invalid `concurrency` clamps rather than fails**, forced by `mapFailuresAsync` returning `Promise<string[]>` with no `Result` to fail into.

New **§ 7 — the in-repo migration is deliberately deferred**, with the reason recorded: it's a **behaviour change, not a refactor**. `HybridRetriever`, the prompt-assist observer fan-out, and the bcp47 loader are unbounded today and would pick up `DEFAULT_RESULT_CONCURRENCY` (8). That's the right default, but whether each site wants it — versus `Number.POSITIVE_INFINITY` to preserve current behaviour — is a per-site judgement that shouldn't ride along in a mechanical sweep.

## `fetch-primitive-threat-model.md`

Marked **partially implemented**, naming what shipped (#592, #594 via #597) and what's in flight (S2b redirect walk, S3 browser/retry/docs).

---

### DIFF
--- a/.claude/project/async-result-family-design.md
+++ b/.claude/project/async-result-family-design.md
@@ -368,8 +368,8 @@ correct `(item) => ...` callback that ignores its index, trading a rare bug for 
 ## 7. Follow-up: migrating the in-repo call sites
 
 The three hand-rolled sites in § 2(d) — `HybridRetriever`, the prompt-assist observer fan-out,
-the bcp47 registry loader — are **not** migrated. Left deliberately as a separate stream,
-because migration is a **behaviour change, not a refactor**: those sites are unbounded today
+the bcp47 registry loader — are **not** migrated. They are left deliberately as a separate
+stream, because migration is a **behaviour change, not a refactor**: those sites are unbounded today
 and would pick up `DEFAULT_RESULT_CONCURRENCY` (8). That is the right default, but whether each
 site wants it — versus `Number.POSITIVE_INFINITY` to preserve current behaviour — is a
 per-site judgement about its workload, and should not ride along in a mechanical sweep.
--- a/.claude/project/async-result-family-design.md
+++ b/.claude/project/async-result-family-design.md
@@ -1,6 +1,8 @@
 # The async Result family — design
 
-**Status:** design only. No implementation.
+**Status:** **implemented and merged** — #596. This document is retained as the rationale
+record, not as a statement of pending work. Where the implementation diverged from what is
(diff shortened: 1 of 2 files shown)


ErikFortune:
Closing — superseded, content preserved.

Both doc-status corrections (and the sentence-fragment fix from the Copilot pass) have been folded into the `safer-fetch-s3` branch at `5abe19ec6`, so they ship with the stream that finishes the work instead of landing as a standalone docs commit on `release`.

That was the right call for a reason beyond tidiness: this PR only existed because two design docs still announced themselves as unimplemented after their streams landed — drift that happens precisely *because* docs ship separately from the code that makes them true. Merging it as its own commit would have treated the symptom.

One correction found while folding it in: this PR's status line was **right** and the S3 brief I wrote was **wrong**. #594 is S1 (core) and #592 is S2a (address classification), not the reverse. The design doc's own in-body annotations (`Corrections from implementation (S2a, PR #592)`, `Further corrections from implementation (S1, PR #594)`) settle it. The brief and the streams ledger have been fixed to match.

Nothing here is lost; the branch carries all of it.


---

PR #599, opened 2026-08-04
feat(safer-fetch): DNS resolution, per-hop address guard, redirect walk (S2b)

ErikFortune:
Stream **S2b** of the `safer-fetch` workstream. Joins S1's core to S2a's address classifier and makes the guard load-bearing.

Branches from and targets **`release`** — the brief said `integration/safer-fetch`, but that branch has been squashed to `release` and no longer exists. **Do not merge**; this is for review.

## What ships

**`blockPrivateNetworks()` — the DNS-resolving address guard (Node only).**
Classifies `url.hostname`, never raw URL text. Resolves only when the hostname is not already an IP literal — a `classifyAddress` failure means "not a literal, so resolve it", never "block". Delegates every resolved address to S2a's reject-if-any policy **unchanged**; no classification logic is reimplemented. The `HostResolver` seam is injectable, so no test in the suite touches the network. `node:dns` rejects on `ENOTFOUND`/`EAI_AGAIN`, so the default resolver wraps it in `captureAsyncResult` — a rejected lookup is a `Failure`, never a throw out of an API documented to always return a `Result`. Lives in its own module so the packlet's browser barrel stays free of `node:dns`.

`lookup` rather than `resolve4`/`resolve6`: `lookup` goes through the OS resolver — the same path the connect takes — so it sees `/etc/hosts`. `resolve4` would miss a hosts-file entry pointing an allowlisted name at `127.0.0.1`, which is a bypass rather than a curiosity.

**`redirectPolicy: 'validate-each-hop'` — the walk.**
`redirect: 'manual'`, the full address guard on **every** hop before any connection, `Location` resolved against the hop that sent it. The chain is passed whole; hop 0 is not a special case. `IRequestHop.status` is populated; `connectedAddress` is recorded once a guard pins and the transport accepts (sound because a transport that cannot honor a pin must fail rather than connect by hostname).

Plus: `maxRedirects` (default 5), whole-chain loop detection, Fetch-standard method/body rewriting (301/302 convert `POST`→`GET` and drop the body; 303 converts any non-`GET`/`HEAD`; 307/308 preserve both — a dropped body takes `Content-Type`/`Content-Length` with it), and monotonic credential stripping.

**Credential stripping is monotonic, and that is the load-bearing part.**
Headers carry forward from the request actually sent on the previous hop, and the origin comparison is against the hop that redirected. Deriving each hop's headers from the caller's original set and comparing to the caller's original origin finds the last hop of `A`(auth) → `B` → `A` same-origin with `A` and hands the token back, after `B` has already watched the chain go past. It reviews as correct and it is a real disclosure. Dedicated test: `A → B → A does NOT re-attach credentials on the final hop`. Matching is case-insensitive — a request guard returning a replacement request spelled `Authorization` must not slip past a set keyed on `authorization`; also tested.

`sensitiveHeaders` and `maxRedirects` are added to `ISaferFetchOptions` now, with the behaviour behind them — S1 deliberately left them off rather than ship an option that does nothing.

## Naming end state (the decision I was asked to own)

**Unsuffixed factory ⇒ `IAddressGuard`. `Policy`-suffixed factory ⇒ `IAddressPolicy`.**

- `allowAnyAddress()` / `blockPrivateNetworks()` → guards (what `saferFetch*` takes)
- `allowAnyAddressPolicy()` / `blockPrivateNetworksPolicy()` → pure synchronous policies

So the policy factory is renamed `blockPrivateNetworks` → `blockPrivateNetworksPolicy`, and the unsuffixed name moves to the guard layer. Rationale: design § 6.1 hands callers `blockPrivateNetworks()` as the thing they pass to an entry point, so the clean name belongs to the guard; and the rule now generalizes rather than being a per-symbol accident. Diagnostic `name` strings track the factory names (`'blockPrivateNetworksPolicy(allowLoopback)'`, `'blockPrivateNetworks(allowLoopback)'`). Breaking, but only on an unreleased surface within the same alpha.

## Where the design was wrong or silent once I hit the code

1. **`headersTimeoutMs` is documented "per attempt" but was retired for the whole call.** `DeadlineWatch.headersReceived()` clears the headers timer permanently. Correct with one attempt; with a walk, hop 2+ had **no headers deadline at all** — a chain whose second host simply never answers would hang for the full 30s overall budget instead of failing as `timeout.phase === 'headers'`. Added `attemptStarted()`, called at the top of each hop (before the address guard, so guard/DNS time stays inside the budget per § 3.6). This is a defect my change would have introduced, not a pre-existing one. Tested.
2. **Default `redirectPolicy` stays `'reject'`, not `'validate-each-hop'`.** § 6.2 names `'validate-each-hop'` the *Node* default, but S1 shipped one core serving both barrels and there is no runtime split to hang a per-runtime default on. `'reject'` is the only mode with identical semantics on both, and following a chain into hosts the caller never named is a posture worth spelling at the call site — same polarity as `addressGuard` having no default. Flagging it as a deliberate deviation; happy to flip it if you'd rather match § 6.2 literally.
3. **307/308 with an unreplayable body is not reachable.** § 4 step 3c says a stream body makes 307/308 unfollowable. `ISaferFetchOptions.body` is `string | Uint8Array` — always replayable — so there is no case to fail. No code, noted here so the design's line is not read as unimplemented.
4. **Loop detection fires one hop earlier than the brief's example implies.** `A→B→A→B` is caught when `B` names `A` again (2 requests), not on the fourth hop. Strictly better; the test asserts the earlier stop.

## Failure taxonomy — collapsed rather than widened (L4)

No new `FetchFailureReason` variants. Three follow-time conditions share `'redirect-rejected'`: policy is `'reject'`, no usable `Location`, or the target repeats a URL already in the chain. Splitting them would buy a caller almost nothing and would widen the scanning oracle; the two follow-time cases are facts about the chain the *redirecting server* produced, not about the network behind this process. "The chain got too long" stays a separate question as `'too-many-redirects'`, whose message names no URL. An unparseable or non-`http(s)` `Location` reuses `'invalid-url'`.

## Not in scope / unchanged

Pinned-connect: `IGuardVerdict.pinnedAddress` stays `undefined`, `platformFetchTransport` still fails rather than ignore a pin, and **DNS rebinding remains a stated open limit** — no TSDoc implies otherwise. The guard applies no host/port allowlist and no `allowInsecureHttp`; § 13 L6's example calls for those, and they are additive follow-ons (see below). Retry, the browser packlet, README guarantee tables and `LIBRARY_CAPABILITIES.md` are S3's.

## Testing

Everything through the mock `IFetchTransport`; DNS through the injected `HostResolver`. No live server and nothing that could reach `169.254.169.254`. Two `nodeHostResolver` tests do exercise the real resolver but are answered locally — `localhost` from the hosts file, and a name past the 255-octet limit rejected by `getaddrinfo` before it asks anyone.

Beyond the brief's required list: a **`302` → `169.254.169.254` blocked at the hop with `expect(requestedUrls(transport)).toEqual([START])`** — asserted on the transport, not just the returned failure, because a failure raised after the connect looks identical to the caller and would still have hit the metadata service. Also: the guard sees the whole chain with per-hop statuses; a hop resolving to one public + one private address is refused; a scheme-downgrade and a port change each count as cross-origin; a request guard rejecting or retargeting a later hop; the per-attempt headers deadline.

## Gates

- `rushx build` ✅ / `rushx lint` ✅ / `rushx test` ✅ — **100% statements, branches, functions, lines** across every safer-fetch file (and the package).
- `rushx fixlint` run before the final commit; `rush prettier` ran on the commit hook.
- `etc/ts-extras.api.md` regenerated. `ae-unresolved-link` "does not have an export" count is **84 before and 84 after** — no new warnings of that class.
- Rush change file added (`minor`, since it carries the unreleased-surface rename).

## Review-loop discipline

**No `code-reviewer` agent-spawn tool was available in this session**, so layer 1 was a **manual pass against `CODE_REVIEW_CHECKLIST.md`** rather than the agent. An independent pass may be wanted; flagging for the orchestrator to commission.

What the manual pass found and resolved:

- **P1 — none outstanding.** No `any`, no manual-type-check-plus-cast, no double casts in the diff; every fallible operation returns `Result`/`DetailedResult`; no `orThrow` in business logic. The `node:dns` rejection path was the obvious candidate for the "un-captured promise escaping as a throw" class (#594's body-read loop) and is wrapped in `captureAsyncResult`, with a test asserting it surfaces as a `Failure`.
- **P2 — the `headersTimeoutMs` per-attempt defect above**, found by walking the deadline lifecycle against its own TSDoc. Fixed with `attemptStarted()` plus a test.
- **P3 — resolved**: replaced five `expect.stringContaining(...) as unknown as string` matcher casts with `expect.objectContaining`, matching the sibling `saferFetch.test.ts` idiom; dropped a redirect-policy test that duplicated an existing one; switched the scripted-transport helper from URL-keyed object literals to entry tuples (the keys are URLs, which no identifier naming convention accepts).

Coverage closure was **not** needed — the scenario-driven tests reached 100% on their own, so no `c8 ignore` directives were added anywhere.

## For S3

- The `redirectPolicy` default is a live question (item 2 above) — worth settling before the README's guarantee table is written.
- `blockPrivateNetworks` takes no `allowHosts` / `allowPorts` / `allowInsecureHttp`. § 13 L6's Ollama reconciliation example uses all three; adding them is purely additive on `IBlockPrivateNetworksGuardOptions` and the guard is the right place. Today the loopback sidecar case works via `{ allowLoopback: true }` alone, since the core permits `http:` and nothing narrows ports.
- The browser barrel exports the widened `SaferFetchRedirectPolicy`. `'validate-each-hop'` in a browser fails as `'redirect-opaque'` — honest, but S3 should decide whether the browser entry points should refuse the mode up front instead.
- `ALWAYS_STRIPPED_HEADERS` and `DEFAULT_MAX_REDIRECTS` are exported and are the right things to cite in the README's guarantee table.

Copilot review loop not yet started.

---

### DIFF
--- a/common/changes/@fgv/ts-extras/safer-fetch-redirect-walk_2026-08-04-00-00.json
+++ b/common/changes/@fgv/ts-extras/safer-fetch-redirect-walk_2026-08-04-00-00.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@fgv/ts-extras",
+      "comment": "SaferFetch: join the core to the address classifier and make the guard load-bearing. Adds blockPrivateNetworks() — a Node-only, DNS-resolving IAddressGuard that classifies url.hostname (resolving it only when it is not already an IP literal) and delegates every resolved address to the reject-if-any policy, with an injectable HostResolver seam so no test touches the network; a rejected lookup becomes a Failure, never a throw. Adds redirectPolicy: 'validate-each-hop', which walks redirects with redirect: 'manual' and runs the full address guard on every hop before any connection, resolving Location against the hop that sent it; plus maxRedirects (default 5), whole-chain loop detection, Fetch-standard method/body rewriting (301/302 convert POST to GET and drop the body, 303 converts any non-GET/HEAD, 307/308 preserve both), and monotonic credential stripping on cross-origin hops (authorization/cookie/proxy-authorization plus the new sensitiveHeaders option, matched case-insensitively) so an A -> B -> A chain never re-attaches the credential. IRequestHop.status is now populated, and connectedAddress is recorded once a guard pins and the transport accepts it. BREAKING (unreleased surface): the policy-layer factory blockPrivateNetworks is renamed blockPrivateNetworksPolicy, freeing the unsuffixed name for the guard callers actually pass to saferFetch* — unsuffixed factories now uniformly return an IAddressGuard and Policy-suffixed ones an IAddressPolicy. Also fixes a defect the walk exposed: headersTimeoutMs is documented as per-attempt but was retired for the whole call by the first hop's headers, so later hops had no headers deadline; it is now re-armed per attempt. Default redirect policy remains 'reject'. DNS rebinding remains an open, documented limit — IGuardVerdict.pinnedAddress stays undefined and platformFetchTransport still fails rather than ignore a pin.",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@fgv/ts-extras",
+  "email": "noreply@anthropic.com"
+}
--- a/libraries/ts-extras/etc/ts-extras.api.md
+++ b/libraries/ts-extras/etc/ts-extras.api.md
@@ -349,6 +349,9 @@ function allowContentTypes(types: ReadonlyArray<string>): Result<IResponseHeader
 const allProviderIds: ReadonlyArray<AiProviderId>;
 
 // @public
+const ALWAYS_STRIPPED_HEADERS: ReadonlyArray<string>;
+
+// @public
 function anthropicEffortToBudgetTokens(effort: NonNullable<IAnthropicThinkingConfig['effort']>): number;
 
 // @public
@@ -378,10 +381,15 @@ function base64UrlNoPadDecode(encoded: string): Result<Uint8Array>;
 // @public
 function base64UrlNoPadEncode(data: Uint8Array): string;
 
+// Warning: (ae-unresolved-link) The @link reference could not be resolved: This type of declaration is not supported yet by the resolver
+//
+// @public
+function blockPrivateNetworks(options?: IBlockPrivateNetworksGuardOptions): IAddressGuard;
+
 // Warning: (ae-unresolved-link) The @link reference could not be resolved: The package "@fgv/ts-extras" does not have an export "IBlockPrivateNetworksOptions"
 //
 // @public
-function blockPrivateNetworks(options?: IBlockPrivateNetworksOptions): IAddressPolicy;
+function blockPrivateNetworksPolicy(options?: IBlockPrivateNetworksOptions): IAddressPolicy;
 
 // @public
 function callProviderCompletion(params: IProviderCompletionParams): Promise<Result<IAiCompletionResponse>>;
@@ -618,6 +626,9 @@ const DEFAULT_HEADERS_TIMEOUT_MS: number;
 const DEFAULT_KEYSTORE_ITERATIONS: number;
 
 // @public
+const DEFAULT_MAX_REDIRECTS: number;
+
+// @public
 const DEFAULT_MAX_RESPONSE_BYTES: number;
 
 // Warning: (ae-unresolved-link) The @link reference could not be resolved: The package "@fgv/ts-extras" does not have an export "IProviderListModelsParams"
@@ -788,7 +799,18 @@ type FetchFailureReason =
     readonly guard: string;
     readonly detail: string;
 }
-/** A redirect status was received under a redirect policy that rejects redirects. */
+/**
+* A redirect status was received that this call would not follow: the policy is `'reject'`,
+* or the response carried no usable `Location`, or following it would revisit a URL already
+* in the chain.
+*
+* @remarks
+* The three are deliberately one kind. Splitting them would buy a caller almost nothing and
+* would widen the scanning oracle this taxonomy already is — and the two follow-time cases
+* are facts about the chain the redirecting server produced, not about the network behind
+* this process. "The chain got too long" is a different question and stays
+* `'too-many-redirects'`.
+*/
 | {
     readonly kind: 'redirect-rejected';
     readonly url: string;
@@ -969,6 +991,11 @@ function hexDecode(encoded: string): Result<Uint8Array>;
 //
 // @public
 function hexEncode(data: Uint8Array): string;
+
+// Warning: (ae-unresolved-link) The @link reference could not be resolved: This type of declaration is not supported yet by the resolver
+//
+// @public
+type HostResolver = (hostname: string) => Promise<Result<ReadonlyArray<string>>>;
 
 // @public
 class HpkeProvider {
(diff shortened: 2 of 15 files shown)


ErikFortune:
## Copilot review loop — closed at 3 rounds

Supersedes "Copilot review loop not yet started" in the description above. Stopped on the finding profile, not the round count: rounds 1 and 2 each surfaced a real defect, and round 3's remainder is one design decision assigned to S3 plus one item dispositioned as intentional.

### Round 1 → fixed in `994ce5506`

`FetchFailureReason.redirect-rejected.url` meant two different things. Two of three construction sites set it to the URL that *issued* the redirect; the loop-detection site set it to the redirect *target*, while its own message narrated from the issuing side.

Made consistent on "the URL that issued the redirect", with the target named in the message so nothing is lost, and **the field's contract documented on the type** — it had none, which is what let three sites drift apart.

The tell was in the test: it asserted `url: a`, which reads as "the original URL" but was really the repeated target (the issuing hop was `b`). A test that looks right for the wrong reason is why this survived.

Also clarified that under `'reject'` the guarantee is equivalent across runtimes but the failure *kind* is not — Node yields `'redirect-rejected'`, a browser yields `'redirect-opaque'` with no status to report.

### Round 2 → fixed in `c2b7e0588`

`sensitiveHeaderSet` called `.toLowerCase()` on each entry, so a JavaScript caller passing a non-string **threw out of an API contracted to only ever return a `Result`** — the same escape class as the #594 body-read loop.

`_resolveCallOptions` already defends against exactly this for `redirectPolicy`, with a comment saying why. `sensitiveHeaders` was the one option that skipped the check.

**Rejecting rather than coercing, deliberately.** Copilot suggested `String(...)`; on the option that decides which headers are stripped cross-origin, `String(123)` is a header name that silently never matches — a caller who fumbled the type would believe a credential was being stripped while it was carried across origins. Non-arrays are rejected too (spreading a bare string enumerates it into characters), and the check uses `findIndex` rather than `find`, since an array containing `undefined` is itself invalid and `find` cannot distinguish that from "no offender".

Second finding: the browser barrel claimed everything runtime-agnostic ships from both barrels, but `classifyAddress` and the pure policies ship from the Node barrel only. Corrected the wording rather than widening the browser export surface — that is a design call belonging to the stream that owns the browser packlet.

### Round 3 → both dispositioned, documented in `b8418ded3`

**Loop detection vs. guard normalization — real, assigned to S3.** `completed` holds guard-cleared URLs; `to` is the raw resolved `Location` the guard has not seen yet. A guard that normalizes (`IGuardVerdict.url` explicitly permits it) can produce a stored URL a later raw target does not match, deferring a repeat to `maxRedirects`.

Not reachable with any shipped guard — neither `allowAnyAddress` nor `blockPrivateNetworks` normalizes, and the WHATWG parser already lowercases hosts and punycodes IDNs on both sides. Bounded when it is reachable: every hop is still fully address-guarded, so it costs a longer walk, not a bypass.

The correct fix compares like with like — check after the guard clears the hop — which means splitting the guard step out of `_connect`. **That is S3's**, added to the handoff list. Special-casing trailing dots here would be worse than leaving it: this layer does not own normalization and cannot anticipate a custom guard, so it would advertise a completeness it would not have. Documented at the site.

**Guard vs. policy message prefixes — intentional, now documented.** DNS failures read `blockPrivateNetworks(...)`; policy rejections read `blockPrivateNetworksPolicy(...)`. That is the two-layer split this PR's rename established: each message names the layer that actually rejected, so a reader can tell whether DNS failed or an address was classified and refused. `blocked-by-guard.guard` names the guard either way. Flattening both would read as more consistent and carry strictly less information.

### Gates (re-run after every round)

`rushx build` ✅ · `rushx lint` ✅ · `rushx test` ✅ — **2,533 passing, 0 failures, 100% statements/branches/functions/lines**. No `c8 ignore` directives anywhere in the packlet. api.md carries doc-only deltas; no signature changes beyond those already described.

Round 1's gates were re-run after the prettier commit hook reformatted a file post-verification, since the original run no longer covered what was pushed.

### Added to "For S3"

- Move redirect loop detection to after the address guard clears the hop, so both sides of the comparison are guard-cleared URLs (round 3 above).

### Still open for human review

Copilot's round-1 assessment stands and I would echo it: this is security-sensitive code — redirect following, credential stripping, DNS/address validation — and warrants a human read regardless of bot verdicts. The `redirectPolicy` default staying `'reject'` is settled (confirmed: keep it; per-runtime default is S3's).


---

PR #601, opened 2026-08-05
feat(safer-fetch): retry, guard-cleared loop detection, and the browser packlet (S3)

ErikFortune:
Closes out the safer-fetch series — design §14 phases 4–6. With this, `.claude/project/fetch-primitive-threat-model.md` is marked **fully implemented**, and every place the code departed from it is recorded in its new **Appendix D**.

## What landed

**1. Retry (design §11)** — `IRetryPolicy`, off by default. Every rule in §11 is implemented and has a test:

- Retryable: `network`, `timeout`, and `http-status` in `{408, 429, 500, 502, 503, 504}`. Never retryable: everything else, including every other 4xx — retrying a guard block cannot succeed, and retrying `too-large` re-downloads up to the cap each time, turning a defense into an amplifier.
- `GET`/`HEAD` only unless `retryNonIdempotent`.
- `Retry-After` honored on 429/503, **clamped to `maxDelayMs`** — it is an attacker-controlled header.
- Exponential backoff with full jitter; the overall deadline is the ceiling, so `{ timeoutMs: 30_000, retry: { attempts: 3 } }` is a thirty-second call.
- **Every attempt re-walks from hop 0** — a full walk, never a resume, never a cached verdict. A cached verdict would give an attacker `N` connects against one check and a retry delay is exactly when a short-TTL rebind lands; a resume is independently wrong because the redirect chain is not stable across attempts. Tests assert the guard sees `[[a], [a,b], [a], [a,b]]` on a retried redirect chain, and that a retry sends the caller's original headers rather than the previous attempt's stripped carry-over.

`DeadlineWatch` now distinguishes an **attempt-scoped** stop (the headers deadline) from a **terminal** one (the overall deadline, the caller's signal). Without it a `timeout` — the failure retry exists for — could never be retried. Found by a test, not by review.

**2. Loop-detection restructure** (inherited from #599 round 3) — moved to after the address guard clears the hop, so both sides of the comparison are guard-cleared URLs. The `KNOWN LIMIT` comment is gone because it no longer applies. The regression test drives a normalizing (trailing-dot-stripping) guard: the repeat is now caught before it is issued instead of running to `maxRedirects`.

**3. Result-chaining pass** — a `_raced` boundary helper folds the stopped/refused/cleared ladder into one chainable `DetailedResult`, and `_clearAddress` splits the guard step out of `_connect` (the same restructure deliverable 2 needed — done once). `_execute`'s `try`/`finally` became `captureAsyncResult`, which also makes a throwing caller-supplied `logger` a `Result` rather than an escaped exception, and the retry loop became a recursive `_runAttempt` — a retry *is* a fresh start, and the self-call keeps "nothing carries over" visible in the shape of the code.

**4. Browser packlet** in `@fgv/ts-web-extras` — the core minus the guard, stating its non-guarantees rather than degrading quietly. `redirectPolicy: 'validate-each-hop'` is refused **at option resolution** naming the runtime, and `addressGuard` stays **required** so a browser call site spells `allowAnyAddress()` and a reviewer can grep it.

**5–6. Docs ship with the code** — the §5.4 guarantee table in both READMEs (reconciled against what actually shipped), the `LIBRARY_CAPABILITIES.md` entry (deliberately **not** in the Result-integration-boundary list — there is no upstream and the opinion is the deliverable), the design status line + Appendix D, and this stream's ledger entry moved to Completed.

**7. Testbed** — a `safer-fetch-guard` scenario driving the Node path against a scripted local `node:http` server on loopback: default posture refusing it, the named opt-in permitting exactly one host and port, the content-type gate, the size cap under a lying `Content-Length`, a guarded redirect walk, retry, and a `302` to `169.254.169.254` refused *at the hop* before any connection.

**Also:** `blockPrivateNetworks` gains `allowHosts` / `allowPorts` / `allowInsecureHttp` and now requires `https:` unless opted out — design §12 listed that default from the start but no shipped guard implemented it (Appendix D-d).

## Open questions

- **OQ-1 — declined the brief's recommendation.** The browser path keeps `redirect: 'manual'` rather than switching to `'error'`. The guarantee is identical either way (a manual redirect in a browser is opaque and is then rejected), but `'error'` rejects with a generic `TypeError` indistinguishable from a DNS failure, so the taxonomy would report `'network'` and lose the fact that a redirect was the cause. For a primitive whose structured taxonomy is itself a deliverable, that is the wrong trade. §5.4's row is restated in both READMEs as "✅ (enforced; surfaces as `'redirect-opaque'`)" — the table and the code agree. Appendix D-a. **This is the one judgment call I would most like a second opinion on**; reversing it is a small change.
- **OQ-2 / OQ-3 — taken as recommended** (refuse `'validate-each-hop'` up front; export the runtime-agnostic classification layer from the browser barrel, with a note that it cannot substitute for the guard). Appendix D-f, D-g.
- **OQ-4 — taken, with one part declined.** The three allowlist options are in and §13 L6's Ollama example is literally runnable, with a test asserting exactly that. §12's `{443}` port **default** was not adopted: it would reject `https://api.example.com:8443/` with a failure that reads as an SSRF block. Appendix D-c.

## Gates

| | `@fgv/ts-extras` | `@fgv/ts-web-extras` | `testbed` |
|---|---|---|---|
| `rushx build` | ✅ | ✅ | ✅ |
| `rushx lint` | ✅ | ✅ | ✅ |
| `rushx test` | ✅ 2,676 | ✅ 532 | ✅ 490 |
| Coverage | ✅ 100% | ✅ 100% | ✅ 100% |

`rushx fixlint` run before the final commit. **No `c8 ignore` directives** anywhere in either packlet — every coverage gap this stream hit was closed by changing the code (the `captureAsyncResult`, the recursive `_runAttempt`, and a helper split that removed a genuinely unreachable branch), never by a directive.

## Review loop

**Layer 1 — `code-reviewer` on the final diff, run *before* chasing 100% coverage.** No P1.

- **P2 (fixed, `05f424b8`)** — `DeadlineWatch._inBodyPhase` was stale across the inter-attempt backoff, so an overall-deadline expiry during a retry sleep reported `timeout.phase: 'body'` while no bytes were in flight. Diagnostics-only, but `phase` is part of the taxonomy's contract. Regression test added.
- **P3 ×3, dispositioned** — a log line that can fire in a narrow race where the retry then does not happen (the line describes intent; the returned failure is correct); `_walk`'s hop loop still reads imperatively (the brief's own exception for loop control flow — the per-hop helpers it calls do chain now); and a note that OQ-1 departed from the brief's recommendation (recorded above and in Appendix D).

The reviewer independently re-derived the retry/guard/deadline interactions and found no cross-attempt leak of a guard verdict, redirect chain, stripped header, or `Retry-After`, and no way for a retry to outlive `timeoutMs`.

The ordering was load-bearing exactly as the brief predicted: the coverage pass that followed found two gaps, and in both the right answer was to change the code rather than add a directive.

**Layer 2 — Copilot loop closed at 2 rounds** on the finding profile. Round 1 found a real scheme hole in `_checkUrl` (`allowInsecureHttp` admitted *any* non-https scheme), a demo step that proved the wrong thing after the https-by-default change, and four `{@link}` references baking `ae-unresolved-link` warnings into the checked-in api.md. Round 2 generated no new comments; its three suppressed items were correct and fixed anyway, and one of them exposed a real weakness in my own fix. **Full record: [the loop comment below](https://github.com/ErikFortune/fgv/pull/601#issuecomment-5188877206).**

## Still wants a human read

This is security-sensitive code — redirect following, credential stripping, DNS/address validation, and now a retry loop that deliberately re-resolves on every attempt. Two bots agreeing is not a substitute. The exit artifact is at `.ai/tasks/active/safer-fetch-s3/result.md`.

https://claude.ai/code/session_019AVrUvn4BDxWUQWCXhs3GW


ErikFortune:
## Copilot review loop — closed at 2 rounds

Stopped on the finding profile, not the round count: round 1 surfaced a real defect plus two accuracy problems, round 2 generated **no new comments** and three suppressed test-tightening items. Those three were correct and are fixed anyway — and one of them exposed a real weakness in my own fix, which is recorded below because it is the more useful half of the story.

### Round 1 → fixed in `2fa29bc9`

**A scheme hole in `_checkUrl`.** The check blocked only when `allowInsecureHttp` was false, so setting the flag admitted *any* non-`https:` scheme — `ftp:`, `file:`, `gopher:`. The flag opts into `http:`, not into "anything that is not https".

Not reachable through the shipped entry points: the core refuses every non-`http(s)` scheme before a guard ever sees a URL. But `blockPrivateNetworks` is a **public export a caller can hold and invoke directly**, and a security check that is only correct because something upstream happens to be correct is one refactor away from being wrong. Now allows `https:` always and `http:` only under the flag, with a test asserting refusal under both settings.

**A demo step that proved the wrong thing.** The testbed's "default posture refuses loopback" step used an `http:` loopback URL. After this PR's https-by-default change, the refusal it demonstrated was the *scheme* check, not loopback classification — the step would have kept passing if loopback classification broke entirely.

It now uses an `https:` loopback URL (the guard decides before any connection, so nothing needs to serve TLS), and plaintext HTTP gets its own step against a loopback-permitting guard, so neither posture can be mistaken for the other.

**Four `{@link browserSaferFetch*}` references** baked `ae-unresolved-link` warnings into the checked-in `ts-web-extras.api.md`, since those functions export as `SaferFetch.*` rather than top-level. Replaced with code spans per the checklist's cross-package rule. The report drops from 19 warnings to the 15 that pre-date this branch.

### Round 2 → no new comments; three suppressed items fixed in `e16207d2`

All three were the same class as round 1's second finding, now aimed at the steps I had just written: a demo step creditable by the wrong refusal. The scenario's guards come from one family and **all** refuse as `'blocked-by-guard'`, so loopback, scheme and port refusals are indistinguishable at the taxonomy level. `refusedByGuard` now requires the guard's own reason to name the rule the step is about.

**The part worth reading.** I first matched that evidence against the whole failure message, and a test immediately showed why that is wrong: the message echoes the refused URL, so `'6379'` was satisfied by `http://127.0.0.1:6379/` itself no matter which rule said no — the step was being credited by its own input. The check now reads `detail.detail`, the guard's reason, and a test drives a refusal naming a *different* rule to assert the step is not credited. Copilot flagged the shape; writing the fix badly first is what surfaced the sharper version of its point.

Also awaits `server.close()` in the scenario teardown — it is asynchronous, and returning while the listener is still shutting down leaves an open handle for whatever runs next in the process.

### Gates (re-run after every round)

`rushx build` ✅ · `rushx lint` ✅ · `rushx test` ✅ — **2,676 / 532 / 490 passing** across `@fgv/ts-extras`, `@fgv/ts-web-extras` and `testbed`, **100% statements/branches/functions/lines in all three**, and **no `c8 ignore` directives** anywhere in either packlet. Both coverage gaps this stream hit were closed by changing the code rather than the measurement — `try`/`finally` became `captureAsyncResult`, the retry loop became a recursive `_runAttempt`, and the last one here removed a genuinely unreachable branch by splitting a helper that was doing two jobs.

### Still open for human review

Echoing #599's standing note, and it applies at least as much here: this is security-sensitive code — an SSRF address guard, redirect following with credential stripping, and now a retry loop that deliberately re-resolves on every attempt. It warrants a human read regardless of bot verdicts. The internal `code-reviewer` pass, the four open questions and their resolutions, and the divergence record are in `.ai/tasks/active/safer-fetch-s3/result.md` and design Appendix D.

One decision worth a second opinion: **OQ-1**. I kept `redirect: 'manual'` on the browser path rather than switching to `redirect: 'error'` as the brief recommended, because `'error'` would report a generic `'network'` failure where `'manual'` reports `'redirect-opaque'`. The guarantee is identical; only the failure reason differs. I restated §5.4's row instead, so the table and the code agree — but the brief framed it the other way, so if you prefer the design's original phrasing the change is small.

---

